PR #240, opened 2018-12-14
New Scroll Directions

staravia:
Use enum for scroll direction instead of boolean


Swan:
You need to fix all of your conflicts before I can merge

PR #275, opened 2018-12-21
Rename "Default Skin" to "Skin Style"

Sublimity:
The purpose of this PR is to refactor the horizontal selector setting for the _misleading_ "Default Skin" setting, which is proposed to be renamed to "Skin Style".

### DIFF
--- a/Quaver.Shared/Config/ConfigHelper.cs
+++ b/Quaver.Shared/Config/ConfigHelper.cs
@@ -231,9 +231,9 @@ namespace Quaver.Shared.Config
         /// <param name="s"></param>
         /// <param name="newVal"></param>
         /// <returns></returns>
-        internal static DefaultSkins ReadDefaultSkin(DefaultSkins defaultSkins, string newVal)
+        internal static SkinStyles ReadSkinStyle(SkinStyles defaultSkins, string newVal)
         {
-            return Enum.TryParse(newVal, out DefaultSkins newDefault) ? newDefault : defaultSkins;
+            return Enum.TryParse(newVal, out SkinStyles newDefault) ? newDefault : defaultSkins;
         }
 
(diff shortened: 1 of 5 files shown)


Sublimity:
From what AiAe told me, custom skins in quaver work by having two folders inside each skin folder, one for a bar skin and one for an arrow skin, which you then can toggle in game via the use of the "Default Skin" setting.

Saying they are default skins is only true for when not using a custom skin, it is better to rename it to skin styles due to the fact that they are different in terms of design, and just calling them default skins makes no sense. The "Misleading" part is just because it might confuse players like it did with me, having two separate options for skins would be confusing without much context or explaining, so yeah adding a tool tip would be handy

Also maybe "Skin Type" instead of "Skin Style"?

Swan:
This isn't how it works. A custom skin is just one folder that contains elements for your skin - whether you want to make it bars or arrows is up to you, but there is **only** one containing skin. 

The selected default skin in-game is what elements which aren't overridden by the custom skin are defaulted to. 

Skin type seems like a better choice, but it could be even more misleading if we decide to add more default skins such as orbs or even other bar/arrow skins.

I still think a tooltip would be the best solution for this, so we can leave it on the backburner if you have no other suggestions.

Sublimity:
Yeah all good, I guess I misinterpreted what I was told about how it worked, adding a tooltip would be a better option in this case 👍

P.S. 
This is why I misinterpreted how skin's we're mean't to work, I took it as one skin supports multiple styles, which then can be inherently changed by changing the "Default Skin" value in settings instead of having to switch to a different skin all together.
image

Sublimity:
Woops, forgot to close this yonks ago.

Will submit a new PR whenever tool-tips are implemented.

PR #298, opened 2018-12-22
Change maximum length for file path after map conversion to match Windows limit

BetaVN:
Length limit changed from 200 to 260 to prevent file naming errors after map conversion.


Quaver.Shared/Converters/Osu/Osu.cs
```diff
@@ -70,7 +70,7 @@ public static void ConvertOsz(string file, string extractDirectory)
                     var fileName = $"{extractDirectory}/{Path.GetFileName(tempFile)}";
 
                     // Make sure the path to the file is less than 200 characters
```
Swan: This comment should be either updated or removed.

BetaVN: Comment fixed

PR #307, opened 2018-12-22
Added random map selection button

### LINKED ISSUE #304: Random map selection
Sometimes even i don't know the map i would like to play

Random map selection with a key or button

Sublimity:
Random map selection button + F2 Button both do the same function in the map selector, they both will randomly select a mapset/difficulty that is not already selected and has two or more mapset/difficulty available when combined with the search bar or just in general.

The reason for making it have two or more map sets/difficulties available is because the client would crash, it would freeze before it posted any stack trace so I have no idea why it would've crashed, might've been an index out of bounds exception ¯\_(ツ)_/¯

Resolves #304


Swan:
Very cool feature. A bit buggy with the scrolling, but that's due to a separate issue with the way the sets are pooled and scrolled too fast. Merged nevertheless. We'll have to handle this separately.

PR #319, opened 2018-12-22
Add pressing rightclick in difficulty selector to go back

### LINKED ISSUE #282: Allow pressing right click during difficulty select in order to return back to the song select
In order to add more fluidity to the game, it would be nice to be able to use the right click button after clicking on a song, where you are looking at the difficulties, as an alternative to the ESC key to go back to song select.
An Alternative solution would be a song select like osu! where clicking a song set doesn't close you off and only show you one directory, but instead opens similar to a folder and indented out, while still showing the rest of the songs in song select

This is ultimately to help make a game that is very fluid in the menu and song selects.

Attached is a reference to osu!'s way of managing song selects after you click a song series, for referencing the alternative solution
image

Sublimity:
Title says it all

Resolves #282

### DIFF
--- a/Quaver.Shared/Screens/Select/SelectScreen.cs
+++ b/Quaver.Shared/Screens/Select/SelectScreen.cs
@@ -107,6 +107,7 @@ namespace Quaver.Shared.Screens.Select
         {
             KeepPlayingAudioTrackAtPreview();
             HandleInput();
+            HandleMousePressRight();
 
             base.Update(gameTime);
         }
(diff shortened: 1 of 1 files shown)


Quaver.Shared/Screens/Select/SelectScreen.cs
```diff
@@ -107,6 +107,7 @@ public override void Update(GameTime gameTime)
         {
             KeepPlayingAudioTrackAtPreview();
             HandleInput();
+            HandleMousePressRight();
```
Swan: This should be located in `HandleInput()`, not here.

Sublimity: Aight I'll change it real quick

Sublimity:
Oof, branch conflict, forgot to merge from upstream.

Issue #291, opened 2018-12-21
Rename BgMaskPadding -> StageReceptorPadding


staravia:
Bg Mask was the old term for stage, and was not changed in config.

staravia:
Can I get the ability to add tags please @Swan

PR #328, opened 2018-12-22
Changed BgMaskPadding to StageReceptorPadding

JaanJah:
See: https://github.com/Quaver/Quaver/issues/291

NB! This change would mean that all custom skins that still use `BgMaskPadding` would not work!


Swan:
A PR should also be opened in [Quaver.Wiki](https://github.com/Quaver/Quaver.Wiki) that addresses this change.

JaanJah:
@Swan I don't know if I am blind, but I can't even find `BgMaskPadding` anywhere in [Quaver.Wiki/Skins/Skin.ini](https://github.com/JaanJah/Quaver.Wiki/blob/master/Skins/Skin.ini/en.md) or rest of the wiki in fact.

Swan:
Doesn't seem to be there. In that case you can add it to the table. I'd like to merge both at the same time however.

JaanJah:
Added to wiki with this PR: https://github.com/Quaver/Quaver.Wiki/pull/3

PR #329, opened 2018-12-22
Fixed sound-effect playing when it shouldn't

Sublimity:
I realised that the *ding* sound effect that plays when going down and up in the volumecontroller overlay plays even when master volume and or effect volume is at 0.


Swan:
I'm pretty sure this is a larger issue with master volume not having any effect on effect volume. This should be fixed framework side.

https://github.com/Quaver/Wobble/tree/master/Wobble/Audio/Samples

Sublimity:
Aight, I'll take a look into it, and also quickly do a test to see if other effect sounds play when master is on 0%

Sublimity:
Seeming this problem is related to Wobble and not the client directly, I have reverted my changes for documentation reasons in here and will be closing this PR.

A new pr will be submitted for Wobble once it's resolved.

PR #331, opened 2018-12-23
Fixed ESC button resuming over chat overlay when paused

### LINKED ISSUE #326: When paused and chat is open, pressing escape just resumes the game.
While this is pretty obvious why it does this, I feel like pressing escape while in the chat should just go to the pause screen since I'm used to hitting escape to get out of chat. Sometimes when the game resumes it catches me off guard. Just a little nitpick that I feel could be a little better, but of course, doesn't have to be there.

Sublimity:
Resolves #326

### DIFF
--- a/Quaver.Shared/Screens/Gameplay/GameplayScreen.cs
+++ b/Quaver.Shared/Screens/Gameplay/GameplayScreen.cs
@@ -35,10 +35,7 @@ using Quaver.Shared.Skinning;
 using Wobble;
 using Wobble.Audio;
 using Wobble.Audio.Tracks;
-using Wobble.Discord;
-using Wobble.Discord.RPC;
 using Wobble.Graphics.Animations;
-using Wobble.Graphics.UI.Dialogs;
 using Wobble.Input;
 using Wobble.Logging;
 using Wobble.Screens;
@@ -364,12 +361,11 @@ namespace Quaver.Shared.Screens.Gameplay
             // The user wants to resume their play.
             else if (IsPaused && (KeyboardManager.IsUniqueKeyPress(ConfigManager.KeyPause.Value) || KeyboardManager.IsUniqueKeyPress(Keys.Escape)))
             {
-                if (DialogManager.Dialogs.Count != 0)
+                if (ChatManager.IsActive)
                 {
-                    DialogManager.Dismiss();
+                    ChatManager.ToggleChatOverlay();
                     return;
                 }
-
 
                 Pause();
                 TimePauseKeyHeld = 0;
--- a/Quaver.Shared/Screens/Gameplay/GameplayScreen.cs
+++ b/Quaver.Shared/Screens/Gameplay/GameplayScreen.cs
@@ -26,6 +26,7 @@ using Quaver.Shared.Graphics.Notifications;
 using Quaver.Shared.Helpers;
 using Quaver.Shared.Modifiers;
 using Quaver.Shared.Online;
+using Quaver.Shared.Online.Chat;
 using Quaver.Shared.Screens.Gameplay.Replays;
 using Quaver.Shared.Screens.Gameplay.Rulesets;
 using Quaver.Shared.Screens.Gameplay.Rulesets.Input;
@@ -37,6 +38,7 @@ using Wobble.Audio.Tracks;
 using Wobble.Discord;
 using Wobble.Discord.RPC;
 using Wobble.Graphics.Animations;
+using Wobble.Graphics.UI.Dialogs;
 using Wobble.Input;
 using Wobble.Logging;
 using Wobble.Screens;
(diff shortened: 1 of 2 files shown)


Quaver.Shared/Screens/Gameplay/UI/PauseScreen.cs
```diff
@@ -135,9 +139,34 @@ public override void Update(GameTime gameTime)
...
 
+            HandleInput();
+
             base.Update(gameTime);
         }
 
+        /// <summary>
+        ///     Handles all input for the screen.
+        /// </summary>
+        private void HandleInput()
```
Swan: This shouldn't be handled here. It should be done in GameplayScreen.

https://github.com/Quaver/Quaver/blob/develop/Quaver.Shared/Screens/Gameplay/GameplayScreen.cs#L262

Quaver.Shared/Screens/Gameplay/UI/PauseScreen.cs
```diff
@@ -135,9 +139,34 @@ public override void Update(GameTime gameTime)
...
+        /// </summary>
+        private void HandleKeyPressEsc()
+        {
+            if (!Screen.IsPaused)
+                return;
+
+            if (KeyboardManager.IsUniqueKeyPress(Keys.Escape))
+                return;
+            
+            if (ChatManager.IsActive)
```
Swan: Use `DialogManager.Dialogs.Count != 0`. This will handle the case of all dialog screens - not just the chat.

When moving it over to GameplayScreen, you can have this check at the top of the HandleInput method.

Quaver.Shared/Screens/Gameplay/UI/PauseScreen.cs
```diff
@@ -146,9 +175,12 @@ public void Activate()
             ClearTransformations();
 
             Background.Animations.Add(new Animation(AnimationProperty.Alpha, Easing.Linear, 0, 1, 400));
-            Continue.Animations.Add(new Animation(AnimationProperty.X, Easing.OutExpo, -Continue.Width, GetActivePosX(Continue),  400));
-            Retry.Animations.Add(new Animation(AnimationProperty.X, Easing.OutExpo, -Retry.Width, GetActivePosX(Retry), 400));
-            Quit.Animations.Add(new Animation(AnimationProperty.X, Easing.OutExpo, -Quit.Width, GetActivePosX(Quit), 400));
+            Continue.Animations.Add(new Animation(AnimationProperty.X, Easing.OutExpo, -Continue.Width,
```
Swan: Not sure if your IDE automatically did this, but we might as well make this better while we're here.

Use `Continue.MoveToX()`. It's a shorter way of doing the same thing. (and for the other ones)

Sublimity: Nah, my ide didn't do anything to the code for the animations, i think the reason it showed up as changed code is due to formatting the class file, which I accidently did.

Sublimity: But yeah, I will change the code out to MoveToX()

Quaver.Shared/Screens/Gameplay/UI/PauseScreen.cs
Sublimity: When dismissing via the DialogManager, should I dismiss only the current screen, or all?

Sublimity:
@Swan Request's completed

Quaver.Shared/Screens/Gameplay/GameplayScreen.cs
```diff
@@ -362,6 +364,13 @@ private void HandlePauseInput(GameTime gameTime)
             // The user wants to resume their play.
             else if (IsPaused && (KeyboardManager.IsUniqueKeyPress(ConfigManager.KeyPause.Value) || KeyboardManager.IsUniqueKeyPress(Keys.Escape)))
             {
+                if (DialogManager.Dialogs.Count != 0)
+                {
+                    DialogManager.Dismiss();
+                    return;
+                }
+
+
```
Swan: Remove the extra line break here.

Swan: Actually, we should probably just handle chat here now that I think about it. Using Dismiss() will completely destroy the chat interface.

Use this instead.

```cs
if (ChatManager.IsActive)
    ChatManager.ToggleChatOverlay();
```

It was what you had originally, but I wasn't thinking :crying_cat_face:

Sublimity:
aaaaaaaaaaaaaa what so handle it in PauseScreen.cs where I originally had it? Or keep it in GameplayScreen, because either works

Quaver.Shared/Screens/Gameplay/GameplayScreen.cs
Sublimity: Aaaaaaaaaaaaaaaaaaaaa what so handle it in PauseScreen.cs where I originally had it? Or keep it in GameplayScreen, because either works

Swan: Keep it in GameplayScreen

PR #332, opened 2018-12-23
Add non-changeable keybind for opening options from certain menu.

Sublimity:
Implements feature that was to be added from https://github.com/Quaver/Quaver/projects/1#card-15662565

Note: When further screens that have an option to open the options are available, those screens should also have code for this same keybind, the only reason for doing it one by one is because the settings dialog shouldn't be opened from just anywhere.

### DIFF
--- a/Quaver.Shared/QuaverGame.cs
+++ b/Quaver.Shared/QuaverGame.cs
@@ -29,6 +29,7 @@ using Quaver.Shared.Scheduling;
 using Quaver.Shared.Screens;
 using Quaver.Shared.Screens.Alpha;
 using Quaver.Shared.Screens.Menu;
+using Quaver.Shared.Screens.Settings;
 using Quaver.Shared.Skinning;
 using Steamworks;
 using Wobble;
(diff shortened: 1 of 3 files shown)


Swan:
I'm not sure about duplicating this code per screen. If anything we should make this input global in [QuaverGame.cs](https://github.com/Quaver/Quaver/blob/develop/Quaver.Shared/QuaverGame.cs) and only available on certain [QuaverScreenTypes](https://github.com/Quaver/Quaver/blob/develop/Quaver.Shared/Screens/QuaverScreenType.cs).

An example of handling this by screen would be how we handle [replay importing](https://github.com/Quaver/Quaver/blob/develop/Quaver.Shared/Database/Maps/MapsetImporter.cs#L93) - where it's only available on certain screens. The same type of logic can apply to this as well.

Sublimity:
Yeah fairo, I'll make those changes later, also sorry for having to make changes to PR's, no sleep f word's me thinking up.

Issue #251, opened 2018-12-17
Arrow Color Snaps On Converts Become Incorrectly Colored


Reyder95:
When playing a convert, typically there are red, blue and other color arrows depending on the beat snap, but on converts, it seems to glitch and makes some colors completely off (as in will make beats that are supposed to be one color, an entirely different color).

Swan:
Fixed in #346

PR #333, opened 2018-12-23
Use Better Beat-Snap Solver

staravia:
The Beat-Snaps should work as intended now. This solver uses less computation.


staravia:
PR regarding issue #251

PR #340, opened 2018-12-23
Added check for submitting score to quaver servers, if the map is only ranked.

### LINKED ISSUE #338: Submitting score on unranked beatmaps
**Describe the bug**
If you finish an unranked beatmap, on the result screen it will show "Submitting Score".

**To Reproduce**
1. Play an unranked beatmap.
2. See the result screen.
3. See error

**Expected behavior**
I think the "Submitting Score" tag should only appear on ranked beatmaps.

**Screenshots**

Sublimity:
The original issue was to do with unranked/notsubmitted scores still attempting to submit to the Quaver's score server.

Resolves #338

### DIFF
--- a/Quaver.Shared/Screens/Result/ResultScreen.cs
+++ b/Quaver.Shared/Screens/Result/ResultScreen.cs
@@ -299,6 +299,10 @@ namespace Quaver.Shared.Screens.Result
             if (OnlineManager.Status.Value == ConnectionStatus.Disconnected)
                 return;
 
+            // Don't submit scores if they aren't ranked, there is no point.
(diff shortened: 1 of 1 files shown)


Swan:
I intentionally didn't add this check because a map's ranked status is only updated when going to the global scores leaderboard. If a user were to import a map while on local scores and play it, it wouldn't be submitted to the server.

A possible workaround would be to cache if the map's ranked status has been checked in the database, and then make a request regardless of the leaderboard they're on.

Swan:
Closing this. I'll have to implement this myself testing w/ a server.

PR #342, opened 2018-12-23
Skinning interface (for the future)

Abriel14:
The idea would be to have a special interface for skinning with for example:
-an overview of the skin in-game so that you can directly see the modifications and on this overview, one'll be able to modify the position of each components of the skin
-an image browser with previews
-a simple image editor with color changing, rotating, changing size, etc
-and more ideas to come...


Sublimity:
Why was this suggestion made as a Pull Request, it should be made as an issue.

Please close this and remake it as an issue.

PR #375, opened 2018-12-25
Scroll Directions Enum instead of Bool for config

staravia:
Current Scroll Directions:
- Down
- Up
- Split

Guaranteed Upcoming Scroll Directions:
- Left
- Right

Other Minor Tweaks:
- Settings Dialogue: use Apply/Close instead of Ok/Cancel

### DIFF
--- a/Quaver.Shared/Screens/Gameplay/Rulesets/Keys/GameplayRulesetKeys.cs
+++ b/Quaver.Shared/Screens/Gameplay/Rulesets/Keys/GameplayRulesetKeys.cs
@@ -6,6 +6,7 @@
 */
 
 using System;
+using System.Collections.Generic;
 using Microsoft.Xna.Framework;
 using Quaver.API.Enums;
 using Quaver.API.Maps;
@@ -18,6 +19,7 @@ using Quaver.Shared.Screens.Gameplay.Rulesets.Input;
 using Quaver.Shared.Screens.Gameplay.Rulesets.Keys.HitObjects;
 using Quaver.Shared.Screens.Gameplay.Rulesets.Keys.Playfield;
 using Quaver.Shared.Screens.Gameplay.Rulesets.Keys.Playfield.Lines;
+using Quaver.Shared.Skinning;
 
 namespace Quaver.Shared.Screens.Gameplay.Rulesets.Keys
 {
@@ -30,23 +32,26 @@ namespace Quaver.Shared.Screens.Gameplay.Rulesets.Keys
         ///     the playfield and the HitObjectManager.
         ///
         ///     We can't intiialize it in Playfield as that gets created first.
+        ///
+        ///     This is a list because multiple scroll directions require multiple Timing Line Managers.
+        /// 
         /// </summary>
-        public TimingLineManager TimingLineManager { get; }
+        public List<TimingLineManager> TimingLineManager { get; } = new List<TimingLineManager>();
 
         /// <summary>
         ///     Dictates if we are currently using downscroll or not.
         /// </summary>
         /// <exception cref="ArgumentOutOfRangeException"></exception>
-        internal static bool IsDownscroll
+        internal static ScrollDirection ScrollDirection
         {
             get
             {
                 switch (MapManager.Selected.Value.Qua.Mode)
                 {
                     case GameMode.Keys4:
-                        return ConfigManager.DownScroll4K.Value;
+                        return ConfigManager.ScrollDirection4K.Value;
                     case GameMode.Keys7:
-                        return ConfigManager.DownScroll7K.Value;
+                        return ConfigManager.ScrollDirection7K.Value;
                     default:
                         throw new ArgumentOutOfRangeException();
                 }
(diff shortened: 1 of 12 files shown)


staravia:
Please close #240 . Error on my part for deleting original fork.

Quaver.Shared/Screens/Gameplay/Rulesets/Keys/GameplayRulesetKeys.cs
```diff
@@ -30,23 +32,26 @@ public class GameplayRulesetKeys : GameplayRuleset
         ///     the playfield and the HitObjectManager.
         ///
         ///     We can't intiialize it in Playfield as that gets created first.
+        ///
+        ///     This is a list because multiple scroll directions require multiple Timing Line Managers.
+        /// 
         /// </summary>
-        public TimingLineManager TimingLineManager { get; }
+        public List<TimingLineManager> TimingLineManager { get; } = new List<TimingLineManager>();
```
Swan: The timing lines on 7k split scroll are broken

img

Quaver.Shared/Screens/Gameplay/Rulesets/Keys/GameplayRulesetKeys.cs
```diff
@@ -58,10 +63,128 @@ internal static bool IsDownscroll
...
-                TimingLineManager = CreateTimingLineManager();
+            {
+                var playfield = (GameplayPlayfieldKeys)Playfield;
+                switch (MapManager.Selected.Value.Qua.Mode)
+                {
+                    case GameMode.Keys4:
+                        {
+                            switch (ConfigManager.ScrollDirection4K.Value)
+                            {
+                                case ScrollDirection.Down:
```
Swan: This code doesn't look good visually. Just use one line or break it into two lines if it's too long.

Quaver.Shared/Screens/Settings/Elements/SettingsScrollDirection4K.cs
```diff
@@ -0,0 +1,53 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Quaver.Shared.Config;
+using Wobble.Bindables;
+
+namespace Quaver.Shared.Screens.Settings.Elements
+{
+    public class SettingsScrollDirection4K : SettingsHorizontalSelector
```
Swan: There is no need to duplicate this code for both 4k and 7k if it's doing the same thing.

Quaver.Shared/Screens/Gameplay/Rulesets/Keys/GameplayRulesetKeys.cs
```diff
@@ -78,109 +78,93 @@ private void InitializeTimingLines()
...
-                                            playfield.ColumnLightingPositionY[2],
-                                            playfield.Width - halfway,
-                                            playfield.Width - halfway
-                                        ));
-                                        break;
-                                    }
-                            }
-                            break;
+                            case ScrollDirection.Down:
+                                TimingLineManager.Add(new TimingLineManager
```
Swan: `TimingLineManager.Add(new TimingLineManager(this, ScrollDirection.Down, playfield.HitPositionOffsets[0], playfield.Width, 0));`

Use one line. The extra line breaks aren't necessary and make it difficult to read. This applies for the other ones as well.

Swan: Instead of doing it like this, why not use object initialization to add the timing lines instead of calling Add each time. It should do the same thing in a much cleaner way.

Quaver.Shared/Screens/Gameplay/Rulesets/Keys/Playfield/GameplayPlayfieldKeys.cs
```diff
@@ -108,28 +98,171 @@ public GameplayPlayfieldKeys(GameplayScreen screen, GameplayRulesetKeys ruleset)
...
 
+        /// <summary>
+        ///     Determine the Scroll Directions for each Lane.
+        /// </summary>
+        private void DetermineScrollDirections()
+        {
+            switch (Ruleset.Map.Mode)
+            {
+                case GameMode.Keys4:
+                    switch (ConfigManager.ScrollDirection4K.Value)
```
Swan: Not a fan of the nested switches. Break it up into separate methods if u have to.

Quaver.Shared/Screens/Gameplay/Rulesets/Keys/Playfield/GameplayPlayfieldKeys.cs
```diff
@@ -108,28 +98,171 @@ public GameplayPlayfieldKeys(GameplayScreen screen, GameplayRulesetKeys ruleset)
...
+
+            for (var i = 0; i < ScrollDirections.Length; i++)
+            {
+                switch (ScrollDirections[i])
+                {
+                    case ScrollDirection.Down:
+                        HitPositionOffsets[i] = ReceptorPositionY[i] + skin.HitPosOffsetY;
+                        break;
+                    case ScrollDirection.Up:
+                        Console.WriteLine(skin.HitPosOffsetY);
```
Swan: Remove the log.

Quaver.Shared/Screens/Gameplay/Rulesets/Keys/GameplayRulesetKeys.cs
staravia: > Instead of doing it like this, why not use object initialization to add the timing lines instead of calling Add each time. It should do the same thing in a much cleaner way.

Can you elaborate what you mean by this? I don't quite understand.

staravia:
I have addressed all the requests from  4 days ago.

Quaver.Shared/Screens/Gameplay/Rulesets/Keys/GameplayRulesetKeys.cs
staravia: I think I understood what you meant and implemented the method differently. Coffee wise, everything looks much cleaner.

Issue #237, opened 2018-12-14
Mention NoSV + NoFail + Autoplay modifiers are unrankable


staravia:
Proposals:
- Add warning when a score will not be submitted when mods are selected or when player pauses
- Make it clear that modifiers are unrankable during selection

JaanJah:
Added: https://github.com/Quaver/Quaver/pull/391

PR #391, opened 2018-12-26
Warning message when choosing unranked modifiers.

JaanJah:
See: https://github.com/Quaver/Quaver/issues/237

### DIFF
--- a/Quaver.Shared/Screens/Select/UI/Modifiers/DrawableModifierBool.cs
+++ b/Quaver.Shared/Screens/Select/UI/Modifiers/DrawableModifierBool.cs
@@ -32,7 +32,6 @@ namespace Quaver.Shared.Screens.Select.UI.Modifiers
                 new DrawableModifierOption(this, "ON", (o, e) =>
                 {
                     if (!ModManager.Mods.HasFlag(Modifier.ModIdentifier))
-                    {
                         ModManager.AddMod(Modifier.ModIdentifier);
                         switch (Modifier.Name)
                         {
@@ -42,7 +41,6 @@ namespace Quaver.Shared.Screens.Select.UI.Modifiers
                                 NotificationManager.Show(NotificationLevel.Warning, "WARNING! Using this mod combination will not submit your score for ranked!");
                                 break;
                         }
-                    }
                 })
             };
 
--- a/Quaver.Shared/Screens/Select/UI/Modifiers/DrawableModifierBool.cs
+++ b/Quaver.Shared/Screens/Select/UI/Modifiers/DrawableModifierBool.cs
@@ -34,11 +34,13 @@ namespace Quaver.Shared.Screens.Select.UI.Modifiers
                     if (!ModManager.Mods.HasFlag(Modifier.ModIdentifier))
                     {
                         ModManager.AddMod(Modifier.ModIdentifier);
-                        if (Modifier.Name == "No Slider Velocities"
-                            || Modifier.Name == "No Fail"
-                            || Modifier.Name == "Autoplay")
+                        switch (Modifier.Name)
                         {
-                            NotificationManager.Show(NotificationLevel.Warning, "WARNING! Using this mod combination will not submit your score for ranked!");
+                            case "No Slider Velocities":
+                            case "No Fail":
+                            case "Autoplay":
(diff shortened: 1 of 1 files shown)


Quaver.Shared/Screens/Select/UI/Modifiers/DrawableModifierBool.cs
```diff
@@ -31,7 +32,15 @@ public override List<DrawableModifierOption> CreateModsDialogOptions()
                 new DrawableModifierOption(this, "ON", (o, e) =>
                 {
                     if (!ModManager.Mods.HasFlag(Modifier.ModIdentifier))
+                    {
                         ModManager.AddMod(Modifier.ModIdentifier);
+                        if (Modifier.Name == "No Slider Velocities"
```
Swan: String comparison isn't a good way to go about it. What if we wanted to change the name of the modifier?

[IGameplayModifier](https://github.com/Quaver/Quaver/blob/develop/Quaver.Shared/Modifiers/IGameplayModifier.cs) has a property that defines if the mod is ranked or not. 

The warning notification *could* go in [ModManager.AddMod()](https://github.com/Quaver/Quaver/blob/develop/Quaver.Shared/Modifiers/ModManager.cs#L54) if you wanted to approach it this way and check against the created instance of the modifier if it's ranked or not.

However, I think that having some sort of indicator in the mod menu itself would be better than a notification.

Quaver.Shared/Screens/Select/UI/Modifiers/DrawableModifierBool.cs
```diff
@@ -31,7 +32,15 @@ public override List<DrawableModifierOption> CreateModsDialogOptions()
                 new DrawableModifierOption(this, "ON", (o, e) =>
```
Swan: It's worth mentioning that when you click the on/off button, the click event fires again. This should preferably be prevented here as well if we're going this route.

Quaver.Shared/Screens/Select/UI/Modifiers/DrawableModifierBool.cs
```diff
@@ -31,7 +32,15 @@ public override List<DrawableModifierOption> CreateModsDialogOptions()
...
                 {
                     if (!ModManager.Mods.HasFlag(Modifier.ModIdentifier))
+                    {
                         ModManager.AddMod(Modifier.ModIdentifier);
+                        if (Modifier.Name == "No Slider Velocities"
+                            || Modifier.Name == "No Fail"
+                            || Modifier.Name == "Autoplay")
+                        {
+                            NotificationManager.Show(NotificationLevel.Warning, "WARNING! Using this mod combination will not submit your score for ranked!");
```
Swan: If we're using a notification, show the name of the unranked modifier(s) rather than saying "this mod combination".

JaanJah:
Warning icon added to the left on unranked modifier. Still need a tooltip or something to tell user what the warning icon means.

Sublimity:
Tooltips aren't implemented yet, I'll make sure to add one to the unranked icon whenever they are implemented.

Quaver.Shared/Screens/Select/UI/Modifiers/DrawableModifier.cs
```diff
@@ -115,6 +118,26 @@ private void CreateModifierDescription() => ModifierDescription = new SpriteText
             UsePreviousSpriteBatchOptions = true
         };
 
+        private void RankedStatusIcon()
+        {
+            if (!Modifier.Ranked)
+            {
+                Modifier.UnrankedSprite = new Sprite()
```
Swan: @JaanJah 

This is good, but the x position isn't centered because the x position is hardcoded. To center it, you can do the following:

```cs
Modifier.UnrankedSprite = new Sprite()
{
    Parent = this,
    Alignment = Alignment.MidLeft,
    Y = ModifierName.Height - 20,
    X = ModifierName.X / 2f,
    UsePreviousSpriteBatchOptions = true,
    Height = 30,
    Width = 30,
    Image = UserInterface.NotificationWarning
};

Modifier.UnrankedSprite.X -= Modifier.UnrankedSprite.X / 2f;
```

It'll give the following result:

img

PR #396, opened 2018-12-27
Added Reload Skin global hotkey

### LINKED ISSUE #267: Refresh Skin Button
When making changes to a skin, you have to load another skin to switch off of it, then choose the one you want to reload if you want to see the changes you made. This is a minor inconvenience but adds up to a lot of time wasted and unnecessary mouse clicks. Easy solution is a reload button that skips straight to reloading only your skin, easy peasy. Would make skinning a lot easier. Pictured is my high effort portrayal to illustrate what I mean
quaver mspain art

Sublimity:
Global hotkey for Reloading of the current skin can only be used in specific screens.

Also, It's annoying how reloading via the Load method in SkinManager cannot be used on separate threads, meaning I just had to use similar code to what is used for reloading the skin in the SettingsDialog.cs class.

Resolves #267

### DIFF
--- a/Quaver.Shared/QuaverGame.cs
+++ b/Quaver.Shared/QuaverGame.cs
@@ -9,6 +9,7 @@ using System;
 using System.IO;
 using System.Linq;
 using System.Reflection;
+using System.Threading;
 using Microsoft.Xna.Framework;
 using Microsoft.Xna.Framework.Input;
 using Quaver.Shared.Assets;
@@ -92,6 +93,11 @@ namespace Quaver.Shared
                 return $@"{assembly.Version.Major}.{assembly.Version.Minor}.{assembly.Version.Build}";
             }
         }
+        
+        /// <summary>
+        ///     The time that the user has requested their skin be reloaded.
+        /// </summary>
+        private long TimeSkinReloadRequested { get; set; }
 
         /// <inheritdoc />
         /// <summary>
@@ -210,6 +216,8 @@ namespace Quaver.Shared
 
             QuaverScreenManager.Update(gameTime);
             Transitioner.Update(gameTime);
+
+            HandleSkinReloading();
         }
 
         /// <inheritdoc />
@@ -370,6 +378,7 @@ namespace Quaver.Shared
         {
             HandleKeyPressF7();
             HandleKeyPressCtrlO();
+            HandleKeyPressCtrlShiftAltR();
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


Quaver.Shared/QuaverGame.cs
```diff
@@ -433,5 +442,51 @@ private void HandleKeyPressCtrlO()
...
+        /// </summary>
+        private void HandleKeyPressCtrlShiftAltR()
+        {
+            // Check for modifier keys
+            if (!((KeyboardManager.CurrentState.IsKeyDown(Keys.LeftControl) || KeyboardManager.CurrentState.IsKeyDown(Keys.RightControl)) &&
+                  (KeyboardManager.CurrentState.IsKeyDown(Keys.LeftShift) || KeyboardManager.CurrentState.IsKeyDown(Keys.RightShift)) &&
+                  (KeyboardManager.CurrentState.IsKeyDown(Keys.LeftAlt) || KeyboardManager.CurrentState.IsKeyDown(Keys.RightAlt))))
+                return;
+            
+            if (!KeyboardManager.IsUniqueKeyPress(Keys.R))
```
Swan: Pretty sure the keybind in osu! S, not R. That should preferably carry over. I thought it wasn't working because the keybind was different 😛

Quaver.Shared/QuaverGame.cs
```diff
@@ -210,6 +216,8 @@ protected override void Update(GameTime gameTime)
 
             QuaverScreenManager.Update(gameTime);
             Transitioner.Update(gameTime);
+
+            HandleSkinReloading();
```
Swan: Why is this called in update? It should happen when the user presses the key combination.

Quaver.Shared/QuaverGame.cs
Sublimity: Well, I didn't want to directly copy the same keybind, but i'll change it up if you really want.

Swan: Might as well. It's just a bit weird since you have to use two hands to reload it (at least for me).

Sublimity: two hands? i only need my pinky, thumb, ring finger and index finger to click the combination

Quaver.Shared/QuaverGame.cs
Sublimity: It's basically the same code as what is used for skin reloading in SettingsDialog.cs

The Load() function in SkinManager causes errors if you run it on a separate thread, and if you have it on the main thread directly after you FadeIn to black, it'll just block until the loading is done. So using the same code as what is in SettingsDialog, that skin reloading method handles smoothly Fading in and Out after the reloading of the skin.

Swan: Gotcha. No worries then. However if it's the exact same functionality then we shouldn't be duplicating code and can just have one method for this used in both cases.

PR #476, opened 2019-01-06
Add randomize modifier

### LINKED ISSUE #402: "Random" modifier
Not related to a problem as far as I know. The Feature request would be adding a Random Modifier to the existing ones.

**Describe the Feature**
Pretty sure you've already know how this modifier works in different games, but the main purpose of this is to randomize the position of the notes without changing the pattern itself. Adding another entry in the modifiers tab with such function should do the work properly.

Sublimity:
Implementation of randomize modifier done through the use of seed generation, which is stored in replays, and is done through modifying the Qua class file. Seeds are stored in replays after gameplay is completed, and are used for when playing back user score replays, to randomize the lanes to that same seed.

Note: The mod also works with converted mapsets.

Requires Quaver/Quaver.API#20

Resolves #402 

(second pr, the old one started getting out of hand)


Quaver.Shared/Modifiers/Mods/ModRandomize.cs
```diff
@@ -0,0 +1,40 @@
...
+using Quaver.API.Enums;
+
+namespace Quaver.Shared.Modifiers.Mods
+{
+    /// <summary>
+    ///     Randomize gameplayModifier. Randomizes the lanes.
+    /// </summary>
+    internal class ModRandomize : IGameplayModifier
+    {
+
```
staravia: remove whitespace

Quaver.Shared/Screens/Loading/MapLoadingScreen.cs
```diff
@@ -101,6 +102,25 @@ private void ParseAndLoadMap()
...
+                // Generate seed and randomize lanes if Randomize modifier is active.
+                if (ModManager.IsActivated(ModIdentifier.Randomize))
+                {
+                    int seed;
+                    
+                    // If loading from a replay.
+                    if (Replay != null)
+                        seed = Replay.RandomizeModifierSeed;
+                    else
+                    {// If loading gameplay.
```
staravia: comment should be on next line

Quaver.Shared/Screens/Loading/MapLoadingScreen.cs
```diff
@@ -101,6 +102,25 @@ private void ParseAndLoadMap()
...
+                        ModRandomize randomizeModifier = (ModRandomize) ModManager.CurrentModifiersList.Find(x => x.ModIdentifier.Equals(ModIdentifier.Randomize));
+                        randomizeModifier.GenerateSeed();
+
```
staravia: remove whitespace here too

Quaver.Shared/Screens/Loading/MapLoadingScreen.cs
staravia: put it before the else statement actually

Issue #485, opened 2019-01-07
Music player still overlapping in v0.13.0


Jolomann:
**Describe the bug**
Spamming music controls still causes song overlap despite the fixes in #448

**To Reproduce**
1. Spam Z X C V while on main menu
2. Enjoy

**Expected behavior**
One song should play at a time

**Video**
https://streamable.com/bk7l4

**Desktop (please complete the following information):**
 - OS: Windows 10

**Additional context**
I did update to v0.13.0

PR #521, opened 2019-02-02
Fix Jukebox Multiple Tracks Bug

staravia:
Issue: https://github.com/Quaver/Quaver/issues/485

I've added a check for each button to make sure LoadingNextTrack is not active. It has totally fixed the issue, and does not cause any other issues.


YaLTeR:
Ideally it should cancel the currently loading track and load the next one because the current fix simply drops the input if it's too fast; I'm not sure if that would be much more complicated to make though.

Swan:
You can achieve this by using a `CancellationToken`. 

It should be implemented in [AudioEngine.LoadCurrentTrack()](https://github.com/Quaver/Quaver/blob/develop/Quaver.Shared/Audio/AudioEngine.cs#L36) . This will fix all cases of where this happens, and we'll be able to remove the locks around the track as well.

staravia:
Pr will be closed because issue will be solved differently

PR #525, opened 2019-02-03
Add 0.?5x rates

YaLTeR:
Needs https://github.com/Quaver/Quaver.API/pull/42


Swan:
@YaLTeR You also need to handle the case of where the user has pressed either [CTRL+- or CTRL++ to change the rate](https://github.com/Quaver/Quaver/blob/develop/Quaver.Shared/Screens/Select/SelectScreen.cs#L292).

YaLTeR:
:white_check_mark:

PR #570, opened 2019-02-21
Use new Qua when test playing in the editor

Swan:
To isolate scoring to that particular section of the map, so the user doesn't have misses for objects before it.

### DIFF
--- a/Quaver.Shared/Screens/Gameplay/GameplayScreen.cs
+++ b/Quaver.Shared/Screens/Gameplay/GameplayScreen.cs
@@ -224,11 +224,11 @@ namespace Quaver.Shared.Screens.Gameplay
         {
             TimePlayed = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds();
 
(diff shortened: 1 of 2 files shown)


Quaver.Shared/Screens/Gameplay/GameplayScreen.cs
```diff
@@ -212,11 +219,22 @@ public class GameplayScreen : QuaverScreen
         /// <param name="scores"></param>
         /// <param name="replay"></param>
         /// <param name="isPlayTesting"></param>
+        /// <param name="playTestTime"></param>
         public GameplayScreen(Qua map, string md5, List<Score> scores, Replay replay = null, bool isPlayTesting = false, double playTestTime = 0)
         {
             TimePlayed = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds();
 
-            Map = map;
+            var testingQua = ObjectHelper.DeepClone(map);
```
YaLTeR: Move this under the if? Doing a deep clone unconditionally seems unnecessary.

Issue #223, opened 2018-12-11
Switching maps too fast wont update background all the time.


AiAe:

Swan:
Resolved in the new UI

PR #584, opened 2019-02-23
Potential Incorrect BG fix

staravia:
Issue: https://github.com/Quaver/Quaver/issues/223


staravia:
Closed. Different implementation here https://github.com/Quaver/Quaver/pull/629

PR #628, opened 2019-02-25
TaskHandler

staravia:
Used for asynchronous tasks.


YaLTeR:
Shouldn't this be in Wobble?

PR #634, opened 2019-02-25
Map deletion

### LINKED ISSUE #336: Right click to delete beatmap
It would be nice to delete beatmaps, sometimes I either grow old of certain maps or just mistakenly download a terrible map and a deletion option would be great. Perhaps when right-clicking the map you get certain options, and one of those could be a deletion option.

Sublimity:
Resolves #336


Quaver.Shared/Screens/ConfirmDialog.cs
```diff
@@ -0,0 +1,240 @@
...
+using Wobble.Graphics.Sprites;
+using Wobble.Graphics.UI.Buttons;
+using Wobble.Graphics.UI.Dialogs;
+using Wobble.Input;
+using Wobble.Logging;
+using Wobble.Window;
+
+namespace Quaver.Shared.Screens
+{
+    public class ConfirmDialog : DialogScreen
```
Swan: This class isn't necessary. We've already got [ConfirmCancelDialog](https://github.com/Quaver/Quaver/blob/develop/Quaver.Shared/Graphics/Dialogs/ConfirmCancelDialog.cs)

Quaver.Shared/Screens/Select/SelectScreen.cs
```diff
@@ -351,6 +363,32 @@ private static void HandleKeyPressTab()
...
+        /// <summary>
+        ///     Handles when the user presses the del key
+        /// </summary>
+        private void HandleKeyPressDel()
+        {
+            if (!KeyboardManager.IsUniqueKeyPress(Keys.Delete) || DialogManager.Dialogs.Count > 0)
+                return;
+
+            // osu! loaded map check to avoid errors due to the map loaded being virtual from osu.
+            if (MapManager.Selected.Value.Game == MapGame.Osu)
```
Swan: Use `!= Quaver` here for all cases.

Quaver.Shared/Screens/Select/SelectScreen.cs
```diff
@@ -351,6 +363,32 @@ private static void HandleKeyPressTab()
...
+            {
+                // Display error message
+                NotificationManager.Show(NotificationLevel.Error,$"This map cannot be deleted due to it being loaded directly from osu!");
```
Swan: Avoid mentioning the names of other games here.

Quaver.Shared/Screens/Select/SelectScreen.cs
Swan: This check should also be moved to the method that handles deletion, so it can be reused.

Quaver.Shared/Screens/Select/SelectScreen.cs
```diff
@@ -542,5 +580,177 @@ public void SelectRandomMap()
...
             }
         }
+
+        /// <summary>
+        ///     Used to start the deleting process.
+        ///
+        ///     woo now AiAe will stop telling me to add this.
+        /// </summary>
+        /// <param name="type"></param>
+        private void DeleteSelected()
```
Swan: Breaks when trying to delete a mapset when its the only one loaded.

Quaver.Shared/Screens/Select/SelectScreen.cs
```diff
@@ -542,5 +580,177 @@ public void SelectRandomMap()
...
+        /// <summary>
+        ///     Determines if the path is a file or directory then deletes it using the appropriate method.
+        /// </summary>
+        /// <param name="path"></param>
+        private void DeletePath(string path)
+        {
+            var attributes = File.GetAttributes(path);
+
+            if (attributes.HasFlag(FileAttributes.Directory))
+                // its a directory!
```
Swan: Unnecessary comments.

Quaver.Shared/Screens/Select/SelectScreen.cs
```diff
@@ -542,5 +580,177 @@ public void SelectRandomMap()
...
+            confirmDelete.Confirmation += (sender, confirm) =>
+            {
+                if (!confirm)
+                    return;
+
+                MapDeletingInProgress = true;
+
+                try
+                {
+                    lock (this)
```
Swan: Reason for the lock? It's the song select screen, so there's no reason at all to lock this.

Quaver.Shared/Screens/Select/SelectScreen.cs
```diff
@@ -542,5 +580,177 @@ public void SelectRandomMap()
...
+
+                        // Dispose of the selected map.
+                        MapManager.Selected?.Dispose();
+
+                        // Run deletion & Reload in the background.
+                        ThreadScheduler.Run(() => DeleteAndReloadMaps(filePath));
+                    }
+
+                    // Finally show confirmation notification
+                    NotificationManager.Show(NotificationLevel.Success, $"Successfully deleted from Quaver!");
```
Swan: Should show the name of the song/map that was deleted.

Quaver.Shared/Screens/Select/SelectScreen.cs
```diff
@@ -542,5 +580,177 @@ public void SelectRandomMap()
...
+            var preservedMapsetIndex = view.MapsetScrollContainer.SelectedMapsetIndex;
+            var preservedDifficultyIndex = view.DifficultyScrollContainer.SelectedMapIndex;
+
+            try
+            {
+                // delet :amgery:
+                DeletePath(path);
+
+                // Reload cache, reload available mapsets.
+                MapDatabaseCache.Load(false);
```
Swan: Instead of doing a full load, just delete it from the current state of the cache.

Quaver.Shared/Screens/Select/SelectScreen.cs
YaLTeR: And while you're at it add the missing space :jerry:

Quaver.Shared/Screens/Select/SelectScreen.cs
Swan: Deleting a difficulty doesn't update the map banner.

Swan: Deleting a mapset when its the only one loaded and then trying to import another results in a thrown exception

Quaver.Shared/Screens/Select/SelectScreen.cs
```diff
@@ -542,5 +573,184 @@ public void SelectRandomMap()
                     break;
             }
         }
+
+        /// <summary>
+        ///     Used to start the deleting process.
+        ///
+        ///     woo now AiAe will stop telling me to add this.
```
YaLTeR: This is unnecessary.

Quaver.Shared/Screens/Select/SelectScreen.cs
```diff
@@ -542,5 +573,184 @@ public void SelectRandomMap()
...
+        /// <param name="type"></param>
+        private void DeleteSelected()
+        {
+            var view = View as SelectScreenView;
+            var type = view.ActiveContainer;
+            var filePath = string.Empty;
+
+            var selectedMapset = AvailableMapsets[view.MapsetScrollContainer.SelectedMapsetIndex];
+
+            if (AvailableMapsets.Count == 0)
```
YaLTeR: This check should be above the previous statement.

Quaver.Shared/Screens/Select/SelectScreen.cs
```diff
@@ -542,5 +573,184 @@ public void SelectRandomMap()
...
+                return;
+
+            switch (type)
+            {
+                case SelectContainerStatus.Mapsets:
+                    if (selectedMapset == null)
```
YaLTeR: When can this be null?

Quaver.Shared/Screens/Select/SelectScreen.cs
```diff
@@ -542,5 +573,184 @@ public void SelectRandomMap()
...
+                        return;
+
+                    filePath = Path.Combine(ConfigManager.SongDirectory.Value, selectedMapset.Directory);
+                    break;
+
+                case SelectContainerStatus.Difficulty:
+                    var selectedDifficulty = AvailableMapsets[view.MapsetScrollContainer.SelectedMapsetIndex].Maps[view.DifficultyScrollContainer.SelectedMapIndex];
```
YaLTeR: You already have `selectedMapset`, no need to fetch it the second time.

Quaver.Shared/Screens/Select/SelectScreen.cs
```diff
@@ -542,5 +573,184 @@ public void SelectRandomMap()
...
+
+                    if (selectedDifficulty == null)
```
YaLTeR: Can this be null?

Quaver.Shared/Screens/Select/SelectScreen.cs
```diff
@@ -542,5 +573,184 @@ public void SelectRandomMap()
...
+                        return;
+
+                    filePath = selectedMapset.Maps.Count > 1 ? Path.Combine(Path.Combine(ConfigManager.SongDirectory.Value, selectedMapset.Directory), selectedDifficulty.Path) : Path.Combine(ConfigManager.SongDirectory.Value, selectedMapset.Directory);
```
YaLTeR: Extract `Path.Combine(ConfigManager.SongDirectory.Value, selectedMapset.Directory)` into a variable right before this statement for readability.

Quaver.Shared/Screens/Select/SelectScreen.cs
```diff
@@ -542,5 +573,184 @@ public void SelectRandomMap()
...
+                    break;
+
+                default:
+                    throw new ArgumentOutOfRangeException();
+            }
+
+            var confirmDelete = new ConfirmCancelDialog($"Are you sure you want to delete this {( type == SelectContainerStatus.Mapsets ? "mapset" : type == SelectContainerStatus.Difficulty ? "difficulty" : null )} ? ", (sender, confirm) =>
```
YaLTeR: The second condition is unnecessary as you already throw if it's not one of the two.

Quaver.Shared/Screens/Select/SelectScreen.cs
```diff
@@ -542,5 +573,184 @@ public void SelectRandomMap()
...
+        /// <summary>
+        ///     Used to delete a mapset or difficulty, then reload the available maps.
+        /// </summary>
+        /// <param name="path"></param>
+        private void DeleteAndReloadMaps(string path)
+        {
+            var view = View as SelectScreenView;
+            var type = view.ActiveContainer;
+
+            if (view.MapsetScrollContainer.SelectedMapsetIndex < 0)
```
YaLTeR: No, this is not how you do it. You spawn this in a separate thread and it might very well race with the main thread. Here's a plausible scenario: the user presses delete, confirms and then very quickly switches to another mapset before this thread has had a chance to run, then the other mapset will get deleted and not the one the user intended. You need to pass everything you need (so the actual maps to delete) into this function via its arguments and _NOT ACCESS ANY GLOBAL MUTABLE STATE_.

Quaver.Shared/Screens/Select/SelectScreen.cs
```diff
@@ -542,5 +573,184 @@ public void SelectRandomMap()
...
+                return;
+
+            // Externally loaded map check
+            if (MapManager.Selected.Value.Game != MapGame.Quaver)
```
YaLTeR: Should be in the first function, no need to wait all the way until here to check this.

Quaver.Shared/Screens/Select/SelectScreen.cs
```diff
@@ -542,5 +573,184 @@ public void SelectRandomMap()
...
+            {
+                // Display error message
+                NotificationManager.Show(NotificationLevel.Error,"This map was loaded from an external source, it cannot be deleted.");
```
YaLTeR: while you're moving this, add that missing space after ,

Quaver.Shared/Screens/Select/SelectScreen.cs
```diff
@@ -542,5 +573,184 @@ public void SelectRandomMap()
...
+                        throw new ArgumentOutOfRangeException();
+                }
+
+                MapDatabaseCache.OrderAndSetMapsets();
+
+                if (PreviousSearchTerm.Length > 0)
+                {
+                    // Grab the mapsets available to the user according to their previous search term.
+                    AvailableMapsets = MapsetHelper.SearchMapsets(MapManager.Mapsets, PreviousSearchTerm);
+                    AvailableMapsets = MapsetHelper.OrderMapsetsByConfigValue(AvailableMapsets);
```
YaLTeR: When you remove a mapset the ordering or the search does not change, so it is unnecessary to do the whole thing all over. Besides, isn't this racing with literally everything in the main thread? I suspect that simplifying this to a simple removal and moving to the first function will also provide a proper fix for spamming delete-enter which doesn't drop these inputs because "mapset deletion is in progress".

Quaver.Shared/Screens/Select/SelectScreen.cs
Sublimity: pretty sure when no map is selected, which can happen

Sublimity: same goes for the exact same thing in the difficulty case

Quaver.Shared/Screens/Select/SelectScreen.cs
Sublimity: never fucking noticed this, damn im dumb

Quaver.Shared/Screens/Select/SelectScreen.cs
Sublimity: refer to other conversation regarding this

Quaver.Shared/Screens/Select/SelectScreen.cs
YaLTeR: Something tells me that when no map is selected the line above where you index the array is just gonna throw an exception.

Quaver.Shared/Screens/Select/SelectScreen.cs
Sublimity: Pretty sure @Swan said he wanted this here for reusability, not completely sure though.

Quaver.Shared/Screens/Select/SelectScreen.cs
```diff
@@ -351,6 +364,19 @@ private static void HandleKeyPressTab()
...
         }
 
+        /// <summary>
+        ///     Handles when the user presses the del key
+        /// </summary>
+        private void HandleKeyPressDel()
+        {
+            var view = View as SelectScreenView;
+
+            if (!KeyboardManager.IsUniqueKeyPress(Keys.Delete) || DialogManager.Dialogs.Count > 0 || AvailableMapsets.Count == 0 || view.MapsetScrollContainer.SelectedMapsetIndex == -1)
```
YaLTeR: I think the `AvailableMapsets.Count == 0` is redundant now that there's a check for index = -1?

Quaver.Shared/Screens/Select/SelectScreen.cs
```diff
@@ -536,5 +556,139 @@ public void SelectRandomMap()
...
+
+                        // Restore index
+                        if (selectedMapsetIndex == AvailableMapsets.Count)
+                            view.MapsetScrollContainer.SelectMapset(AvailableMapsets.Count - 1); // set to end
+                        else
+                            view.MapsetScrollContainer.SelectMapset(selectedMapsetIndex);
+                        break;
+
+                    case SelectContainerStatus.Difficulty:
+                        if (selectedMapset.Maps.Count > 1)
```
YaLTeR: Instead of checking this condition in already two places and duplicating all the code from the previous case branch, check it once in the beginning of the function and condition on that instead of `type` + checking in the Difficulty case branch.

Quaver.Shared/Screens/Select/SelectScreen.cs
```diff
@@ -536,5 +556,139 @@ public void SelectRandomMap()
...
+            {
+                case SelectContainerStatus.Mapsets:
+                    filePath = Path.Combine(ConfigManager.SongDirectory.Value, selectedMapset.Directory);
+                    break;
+
+                case SelectContainerStatus.Difficulty:
+                    var mapsetPath = Path.Combine(ConfigManager.SongDirectory.Value, selectedMapset.Directory);
+                    var difficultyPath = Path.Combine(mapsetPath, selectedMapset.Maps[selectedDifficultyIndex].Path);
+
+                    filePath = selectedMapset.Maps.Count > 1 ? difficultyPath : mapsetPath;
```
YaLTeR: See comment below.

Quaver.Shared/Screens/Select/SelectScreen.cs
```diff
@@ -536,5 +556,139 @@ public void SelectRandomMap()
...
+                switch (type)
+                {
+                    case SelectContainerStatus.Mapsets:
+                        selectedMapset.Maps.ForEach(MapDatabaseCache.RemoveMap);
+                        MapManager.Mapsets.RemoveAt(selectedMapsetIndex);
+                        AvailableMapsets.Remove(selectedMapset);
+                        view.MapsetScrollContainer.InitializeWithNewSets();
+
+                        // Restore index
+                        if (selectedMapsetIndex == AvailableMapsets.Count)
```
YaLTeR: `Math.Min(selectedMapsetIndex, AvailableMapsets.Count - 1)`

Quaver.Shared/Screens/Select/SelectScreen.cs
```diff
@@ -536,5 +556,139 @@ public void SelectRandomMap()
...
+                    case SelectContainerStatus.Difficulty:
+                        if (selectedMapset.Maps.Count > 1)
+                        {
+                            MapDatabaseCache.RemoveMap(selectedDifficulty);
+                            MapManager.Mapsets[selectedMapsetIndex].Maps.Remove(selectedDifficulty);
+                            selectedMapset.Maps.Remove(selectedDifficulty);
+                            view.DifficultyScrollContainer.ReInitializeDifficulties();
+
+                            view.MapsetScrollContainer.SelectMap(selectedMapsetIndex,
+                                selectedDifficultyIndex == selectedMapset.Maps.Count
```
YaLTeR: `Math.Min`

Quaver.Shared/Screens/Select/SelectScreen.cs
```diff
@@ -536,5 +556,127 @@ public void SelectRandomMap()
...
+                {
+                    case SelectContainerStatus.Difficulty when selectedMapset.Maps.Count > 1:
+                        MapDatabaseCache.RemoveMap(selectedDifficulty);
+                        MapManager.Mapsets[selectedMapsetIndex].Maps.Remove(selectedDifficulty);
+                        selectedMapset.Maps.Remove(selectedDifficulty);
+                        view.DifficultyScrollContainer.ReInitializeDifficulties();
+
+                        // Restore index.
+                        view.MapsetScrollContainer.SelectMap(selectedMapsetIndex,
+                            Math.Min(selectedDifficultyIndex, selectedMapset.Maps.Count) == selectedMapset.Maps.Count
```
YaLTeR: sleeping helps

Quaver.Shared/Screens/Select/SelectScreen.cs
```diff
@@ -536,5 +556,127 @@ public void SelectRandomMap()
...
+            {
+                case SelectContainerStatus.Mapsets:
+                    filePath = Path.Combine(ConfigManager.SongDirectory.Value, selectedMapset.Directory);
+                    break;
+
+                case SelectContainerStatus.Difficulty:
+                    var mapsetPath = Path.Combine(ConfigManager.SongDirectory.Value, selectedMapset.Directory);
+                    var difficultyPath = Path.Combine(mapsetPath, selectedMapset.Maps[selectedDifficultyIndex].Path);
+
+                    filePath = Math.Min(selectedMapset.Maps.Count, AvailableMapsets.Count - 1) > 1 ? difficultyPath : mapsetPath;
```
YaLTeR: see below was in relation to replacing the condition

Quaver.Shared/Screens/Select/SelectScreen.cs
Sublimity: yes

Quaver.Shared/Screens/Select/SelectScreen.cs
```diff
@@ -536,5 +556,96 @@ public void SelectRandomMap()
...
+                BackgroundHelper.Background?.Dispose();
+
+                // Dispose of the currently playing track.
+                AudioEngine.Track?.Dispose();
+
+                // Run path deletion in the background.
+                ThreadScheduler.Run(() => DeletePath(type == SelectContainerStatus.Difficulty && selectedMapset.Maps.Count > 1 ? difficultyPath : mapsetPath));
+
+                // Remove mapset/difficulty from cache and AvailableMapsets list.
+                if (type == SelectContainerStatus.Mapsets || type == SelectContainerStatus.Difficulty && selectedMapset.Maps.Count == 1)
```
YaLTeR: This condition occurs twice (this line and 3 lines above), which is why I suggested making it into a variable. Please do that.

Quaver.Shared/Screens/Select/SelectScreen.cs
```diff
@@ -536,5 +556,93 @@ public void SelectRandomMap()
...
+
+                // Run path deletion in the background.
+                ThreadScheduler.Run(() => DeletePath(deleteMapset ? mapsetPath : difficultyPath));
+
+                // Remove mapset/difficulty from cache and AvailableMapsets list.
+                if (deleteMapset)
+                {
+                    selectedMapset.Maps.ForEach(MapDatabaseCache.RemoveMap);
+                    AvailableMapsets.RemoveAt(selectedMapsetIndex);
+                    MapManager.Mapsets.Remove(selectedMapset);
```
YaLTeR: This doesn't actually remove the mapset because mapsets in `AvailableMapsets` are different objects than the ones in `MapManager.Mapsets`. This is because searching can select only part of the difficulties in a mapset (so a different `Mapset` object is required).

YaLTeR: Could probably remove by `selectedMapset.Directory`.

Quaver.Shared/Screens/Select/SelectScreen.cs
```diff
@@ -536,5 +556,93 @@ public void SelectRandomMap()
...
+                    view.MapsetScrollContainer.InitializeWithNewSets();
+                    view.MapsetScrollContainer.SelectMapset(Math.Min(selectedMapsetIndex, AvailableMapsets.Count - 1));
+                }
+                else
+                {
+                    MapDatabaseCache.RemoveMap(selectedDifficulty);
+                    selectedMapset.Maps.RemoveAt(selectedDifficultyIndex);
+                    MapManager.Mapsets[selectedMapsetIndex].Maps.Remove(selectedDifficulty);
```
YaLTeR: Could be the same issue as with mapsets. Better to remove by MD5 I guess.

Quaver.Shared/Screens/Select/SelectScreen.cs
```diff
@@ -601,15 +601,15 @@ private void DeleteSelected()
...
+                    selectedMapset.Maps.ForEach(x => MapManager.Mapsets.RemoveAll(y => y.Maps.Exists(z => x.Md5Checksum == z.Md5Checksum)));
                     view.MapsetScrollContainer.InitializeWithNewSets();
                     view.MapsetScrollContainer.SelectMapset(Math.Min(selectedMapsetIndex, AvailableMapsets.Count - 1));
                 }
                 else
                 {
                     MapDatabaseCache.RemoveMap(selectedDifficulty);
                     selectedMapset.Maps.RemoveAt(selectedDifficultyIndex);
-                    MapManager.Mapsets[selectedMapsetIndex].Maps.Remove(selectedDifficulty);
+                    MapManager.Mapsets.RemoveAll(x => x.Maps.Exists(y => y.Md5Checksum == selectedDifficulty.Md5Checksum));
```
YaLTeR: yeah great let's delete the entire mapset instead of just 1 dif

Quaver.Shared/Screens/Select/SelectScreen.cs
```diff
@@ -601,15 +601,15 @@ private void DeleteSelected()
                 {
                     selectedMapset.Maps.ForEach(MapDatabaseCache.RemoveMap);
                     AvailableMapsets.RemoveAt(selectedMapsetIndex);
-                    MapManager.Mapsets.Remove(selectedMapset);
+                    selectedMapset.Maps.ForEach(x => MapManager.Mapsets.RemoveAll(y => y.Maps.Exists(z => x.Md5Checksum == z.Md5Checksum)));
```
YaLTeR: `MapManager.Mapsets.RemoveAll(x => x.Directory == selectedMapset.Directory)`

Quaver.Shared/Screens/Select/SelectScreen.cs
YaLTeR: 1. find mapset in MapManager.Mapsets with Directory == selectedMapset.Directory
2. in that particular mapset remove difficulty with md5 == selectedDifficulty.md5

Issue #636, opened 2019-02-25
Using the timeline shouldn't play the map


staravia:
It's really annoying to have it automatically play after navigation

jakads:
it also loops right back to start once it hits the end
shouldn't do that either

PR #637, opened 2019-02-25
Editor Progress Bar automatically play when necessary

staravia:
When the player drags on the progress bar, the audio will only play if it was in a playing state before dragging. Otherwise it will remain paused.

It works exacrly like how youtube handles it.

issue https://github.com/Quaver/Quaver/issues/636


jakads:
it also loops right to start once it hits the end
it shouldn't be like that, it should stay at the end

Swan:
@staravia needs to be rebased.

 Furthermore, the same functionality should be the same for the [VisualizationGraphContainer](https://github.com/Quaver/Quaver/blob/develop/Quaver.Shared/Screens/Editor/UI/Graphing/EditorVisualizationGraphContainer.cs#L222)

PR #654, opened 2019-02-25
Concatenate duplicate configuration keys

### LINKED ISSUE #305: Having two or more of the same key in ini files results in a thrown exception

Sublimity:
Resolves #305


Swan:
Same should be the case for skin.ini as well.

Sublimity:
Someone did point that out, I'll implement it when I have some free time

PR #655, opened 2019-02-26
Replay instant exit

Sublimity:
~~There isn't really any point having the pause menu for replays, so why not just remove it and go directly back to the select screen.~~

#813


YaLTeR:
:thinking: why is the key-related condition different from the one right below :thinking:

like, why not just put `if (InReplayMode) Exit()` right there?

Sublimity:
@YaLTeR im blind, that's why :jerry:

YaLTeR:
Maybe like
```
if (long check for keyboard)
{
    if (replay) { exit and return }
    if (paused) { do that thing it does if paused }
}

PR #687, opened 2019-03-07
Fade audio on results screen transitions

### LINKED ISSUE #527: Audio doesn't fade in from zero volume after a song ends
When finishing a chart and going back to the select screen the audio starts off really loud. Either it really starts off louder than it should or maybe not, either way it should have some sort of a fade from lower volume like it does on other screen transitions.

### LINKED ISSUE #638: Song volume isn't affected during transitions
https://youtu.be/H8egNWTwaNc
lol

Swan:
Fixes #527 
Fixes #638

### DIFF
--- a/Quaver.Shared/Screens/Result/ResultScreen.cs
+++ b/Quaver.Shared/Screens/Result/ResultScreen.cs
@@ -30,6 +30,7 @@ using Quaver.Shared.Database.Scores;
 using Quaver.Shared.Discord;
 using Quaver.Shared.Graphics.Backgrounds;
 using Quaver.Shared.Graphics.Notifications;
+using Quaver.Shared.Graphics.Transitions;
 using Quaver.Shared.Helpers;
 using Quaver.Shared.Modifiers;
 using Quaver.Shared.Modifiers.Mods;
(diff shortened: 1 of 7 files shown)


Quaver.Shared/Screens/Result/ResultScreen.cs
```diff
@@ -362,7 +363,16 @@ public void ExitToMenu()
...
                 return;
 
-            Exit(() => new SelectScreen());
+            Exit(() =>
+            {
+                if (AudioEngine.Track != null)
+                {
+                    lock (AudioEngine.Track)
+                        AudioEngine.Track.Fade(10, 300);
+                }
```
YaLTeR: Forgot to replace this one with a function call

Quaver.Shared/Screens/Result/ResultScreen.cs
```diff
@@ -161,6 +161,9 @@ public ResultScreen(Replay replay)
         /// </summary>
         public override void OnFirstUpdate()
         {
+            if (!AudioEngine.Track.IsDisposed && AudioEngine.Track.IsPlaying)
+                AudioEngine.Track.Pause();
```
YaLTeR: I don't think this is the correct thing to do. If the song is still playing after the map is done, let it play until the end. Besides, this also abruptly pauses the song when opening a score in song select.

PR #694, opened 2019-03-09
Offset Calibrator

### LINKED ISSUE #481: Add offset wizard
It would be great to have an offset wizard similar to Etterna where you can press a key to the beat to automatically set the global audio offset.

Swan:
Closes #481

### DIFF
--- a/Quaver.Shared/Screens/Gameplay/GameplayScreen.cs
+++ b/Quaver.Shared/Screens/Gameplay/GameplayScreen.cs
@@ -891,8 +891,8 @@ namespace Quaver.Shared.Screens.Gameplay
                 var samples = new List<double>(stats.Select(x => (double) x.HitDifference).ToList());
                 var stdDev = samples.StandardDeviation();
 
-                if (Math.Abs(stdDev) < Ruleset.ScoreProcessor.JudgementWindow[Judgement.Good])
-                    offset = (int) samples.Average() + ConfigManager.GlobalAudioOffset.Value;
+                if (stdDev < Ruleset.ScoreProcessor.JudgementWindow[Judgement.Good])
+                    offset = (int) -samples.Average() + ConfigManager.GlobalAudioOffset.Value;
             }
 
             DialogManager.Show(new OffsetConfirmDialog(this, offset));
--- a/Quaver.Shared/Screens/Gameplay/GameplayScreen.cs
+++ b/Quaver.Shared/Screens/Gameplay/GameplayScreen.cs
@@ -894,7 +894,7 @@ namespace Quaver.Shared.Screens.Gameplay
                 var stdDev = samples.StandardDeviation();
 
                 if (stdDev < Ruleset.ScoreProcessor.JudgementWindow[Judgement.Good])
-                    offset = (int) -samples.Average() + ConfigManager.GlobalAudioOffset.Value;
+                    offset = (int) samples.Average() + ConfigManager.GlobalAudioOffset.Value;
 
                 if (!Exiting)
                     DialogManager.Show(new OffsetConfirmDialog(this, offset));
--- a/Quaver.Shared/Screens/Gameplay/GameplayScreen.cs
+++ b/Quaver.Shared/Screens/Gameplay/GameplayScreen.cs
@@ -14,6 +14,7 @@ using Microsoft.Xna.Framework.Input;
 using Quaver.API.Enums;
 using Quaver.API.Helpers;
 using Quaver.API.Maps;
+using Quaver.API.Maps.Processors.Scoring.Data;
 using Quaver.API.Replays;
 using Quaver.Server.Common.Enums;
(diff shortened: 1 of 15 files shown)


Quaver.Shared/Screens/Gameplay/GameplayScreen.cs
```diff
@@ -837,5 +876,26 @@ public override UserClientStatus GetClientStatus()
...
+            if (stats.Count != 0)
+            {
+                var samples = new List<double>(stats.Select(x => (double) x.HitDifference).ToList());
+                var stdDev = samples.StandardDeviation();
+
+                if (Math.Abs(stdDev) < Ruleset.ScoreProcessor.JudgementWindow[Judgement.Good])
+                    offset = (int) samples.Average() + ConfigManager.GlobalAudioOffset.Value;
+            }
+
+            DialogManager.Show(new OffsetConfirmDialog(this, offset));
```
YaLTeR: Should be inside both ifs I think.

Quaver.Shared/Screens/Gameplay/GameplayScreen.cs
```diff
@@ -837,5 +876,26 @@ public override UserClientStatus GetClientStatus()
...
+
+            var offset = 0;
+
+            // Calculate new offset
+            if (stats.Count != 0)
+            {
+                var samples = new List<double>(stats.Select(x => (double) x.HitDifference).ToList());
+                var stdDev = samples.StandardDeviation();
+
+                if (Math.Abs(stdDev) < Ruleset.ScoreProcessor.JudgementWindow[Judgement.Good])
```
YaLTeR: Standard deviation is always >= 0, no need to abs.

Quaver.Shared/Screens/Gameplay/GameplayScreen.cs
```diff
@@ -837,5 +876,26 @@ public override UserClientStatus GetClientStatus()
...
+                    offset = (int) samples.Average() + ConfigManager.GlobalAudioOffset.Value;
```
YaLTeR: Shouldn't this be `(int) -samples.Average()`?

Quaver.Shared/Screens/Gameplay/UI/Offset/OffsetCalibratorTip.cs
```diff
@@ -0,0 +1,77 @@
...
+            {
+                Parent = this,
+                Alignment = Alignment.TopCenter,
+                Image = FontAwesome.Get(FontAwesomeIcon.fa_information_button),
+                Y = 18,
+                Size = new ScalableVector2(24, 24)
+            };
+
+            // ReSharper disable once ObjectCreationAsStatement
+            var text = new SpriteTextBitmap(FontsBitmap.AllerRegular, "Play through the map, and at the end, a new offset will be suggested to you.")
```
YaLTeR:
```suggestion
            var text = new SpriteTextBitmap(FontsBitmap.AllerRegular, "Play through the map, and at the end a new global audio offset will be suggested to you.")
```

Quaver.Shared/Screens/Gameplay/UI/Offset/OffsetConfirmDialog.cs
```diff
@@ -0,0 +1,63 @@
...
+    {
+        private QuaverScreen Screen { get; }
+
+        /// <inheritdoc />
+        /// <summary>
+        /// </summary>
+        /// <param name="screen"></param>
+        /// <param name="offset"></param>
+        public OffsetConfirmDialog(QuaverScreen screen, int offset)
+            : base($"Your suggested offset is: {offset}ms. Would you like to use this?", (o, e) => OnConfirm(screen, offset),
```
YaLTeR:
```suggestion
            : base($"Your suggested offset is: {offset} ms. Would you like to use this?", (o, e) => OnConfirm(screen, offset),
```

Quaver.Shared/Screens/Gameplay/GameplayScreen.cs
Swan: The dialog is the way to exit the screen, so it should always display.

YaLTeR: It should not suggest offset = 0 because that's an error case.

Swan: Auto-restart then?

YaLTeR: If I entered the calibrator by accident, I'll escape out asap without pressing anything, in which case I'd like to quit without any confirmation.

Swan: Made it so it just exits with a notification if it can't calculate it. You can still exit normally w/ escape at anytime.

PR #726, opened 2019-04-19
Add Contributor badge

YaLTeR:


Quaver.Shared/Graphics/Overlays/Chat/Components/Messages/Drawable/ChatBadge.cs
```diff
@@ -87,6 +87,9 @@ public static Texture2D GetIcon(UserGroups groups)
             if (groups.HasFlag(UserGroups.Developer))
                 return FontAwesome.Get(FontAwesomeIcon.fa_code);
 
+            if (groups.HasFlag(UserGroups.Contibutor))
```
Swan: This should be under the supervisor group, not all the way up here. It will give the wrong icon if the user is both a contributor and has another "more important" user group.

PR #739, opened 2019-04-24
Don't Auto Dispose BGM

staravia:
Requires pull from https://github.com/Quaver/Wobble/pull/63
Solves https://github.com/Quaver/Quaver/issues/732


Quaver.Shared/Screens/Editor/EditorScreen.cs
```diff
@@ -476,14 +476,12 @@ private void HandleAudioSeeking()
                 MouseManager.PreviousState.ScrollWheelValue)
             {
                 AudioEngine.SeekTrackToNearestSnap(WorkingMap, Direction.Backward, BeatSnap.Value);
-                SetHitSoundObjectIndex();
```
Swan: These calls are not redundant. They are used to reset the index of the HitObject where hitsounds need to be played when seeking through the song. As a result, this actually broke hitsounds because they do not play when seeking backwards. Furthermore, this commit shouldn't be in this PR if you're working on AudioTracks.

PR #743, opened 2019-04-25
Show current map in player cards in chat

YaLTeR:
Check how the text positions look on Windows. If they look off I'll adjust the values so it looks right on Windows (since Linux has the incorrect font).

image


YaLTeR:
(the reason this PR is on hold is because it's unclear whether scroll containers within scroll containers work correctly and there's a high chance they don't, so this is waiting for https://github.com/Quaver/Wobble/issues/58)

PR #750, opened 2019-04-28
Adds setting to hide hit accuracy

### LINKED ISSUE #748: Make it possible to hide Marv/Perf/ect without skinning.
An option to hide certain judgements would be ideal for people who find it distracting.

Example:
- None (Show all judgements)
- Marv Only
- Marv/Perf
- Marv/Perf/Great
- All (Hide all judgements)

nobbele:
Resolves #748

### DIFF
--- a/Quaver.Shared/Config/ConfigManager.cs
+++ b/Quaver.Shared/Config/ConfigManager.cs
@@ -463,6 +463,11 @@ namespace Quaver.Shared.Config
         internal static Bindable<Keys> KeyEditorIncreaseAudioRate { get; private set; }
 
         /// <summary>
+        ///     Whether hit accuracies such as Marv/Perf should be displayed
+        /// </summary>
+        internal static Bindable<bool> ShowHitAccuracy { get; private set; }
+
+        /// <summary>
         ///     Dictates whether or not this is the first write of the file for the current game session.
         ///     (Not saved in Config)
         /// </summary>
@@ -625,6 +630,7 @@ namespace Quaver.Shared.Config
             LobbyFilterFullGame = ReadValue(@"LobbyFilterFullGame", false, data);
             LobbyFilterOwnsMap = ReadValue(@"LobbyFilterOwnsMap", false, data);
             LobbyFilterHasFriends = ReadValue(@"LobbyFilterHasFriends", false, data);
+            ShowHitAccuracy = ReadValue(@"ShowHitAccuracy", true, data);
 
             // Have to do this manually.
             if (string.IsNullOrEmpty(Username.Value))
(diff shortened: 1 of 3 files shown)


Quaver.Shared/Config/ConfigManager.cs
```diff
@@ -462,6 +462,11 @@ public static class ConfigManager
         /// </summary>
         internal static Bindable<Keys> KeyEditorIncreaseAudioRate { get; private set; }
 
+        /// <summary>
+        ///     Whether hit accuracies such as Marv/Perf should be displayed
+        /// </summary>
+        internal static Bindable<bool> ShowHitAccuracy { get; private set; }
```
Swan: The naming for this is very ambiguous. Not sure what it should be at the moment though.

Quaver.Shared/Screens/Gameplay/UI/JudgementHitBurst.cs
```diff
@@ -86,6 +87,9 @@ public void ChangeJudgementFrames(Judgement j)
         /// <param name="j"></param>
         public void PerformJudgementAnimation(Judgement j)
         {
+            // If show hit accuracy is disabled, do nothing
+            if (!ConfigManager.ShowHitAccuracy.Value) return;
```
Swan: If you're hiding only marvelous and perfects, I would expect there to be some sort of condition here that has that. This currently hides all of them.

Quaver.Shared/Screens/Settings/SettingsDialog.cs
```diff
@@ -334,7 +334,8 @@ private void CreateSections()
                     new SettingsBool(this, "Top Lane Cover", ConfigManager.LaneCoverTop),
                     new SettingsBool(this, "Bottom Lane Cover", ConfigManager.LaneCoverBottom),
                     new SettingsBool(this, "Display UI Elements Over Lane Covers", ConfigManager.UIElementsOverLaneCover),
-                    new SettingsBool(this, "Smooth Accuracy Changes", ConfigManager.SmoothAccuracyChanges)
+                    new SettingsBool(this, "Smooth Accuracy Changes", ConfigManager.SmoothAccuracyChanges),
+                    new SettingsBool(this, "Show Hit Accuracy", ConfigManager.ShowHitAccuracy)
```
Swan: A SettingsBool is probably not the best way to do it since it would be better to toggle individual judgements, but this would require having some sort of checkbox dropdown or field with multiple checkboxes which isn't implemented yet. It works for now however if we're only trying to hide marvs & perfs.

Quaver.Shared/Screens/Gameplay/UI/JudgementHitBurst.cs
Swan: Return should be on a new line.

PR #751, opened 2019-04-28
Adds Localization

### LINKED ISSUE #730: Localization Manager
To handle localizable strings/SpriteText with English fallbacks

nobbele:
Resolves #730

Might be better to load via json or something instead of hardcoding the translations? Should relatively easy to implement if so.

### DIFF
--- a/Quaver.Shared/Localization/LocalizedString.cs
+++ b/Quaver.Shared/Localization/LocalizedString.cs
@@ -1,0 +1,14 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Quaver.Shared.Localization
+{
+    /// <summary>
+    ///     Parameter to <see cref="LocalizationManager.Get(LocalizedString, object[])"/>
+    /// </summary>
+    public enum LocalizedString
+    {
+        FPS_IS_NOW_UNLIMITED, FPS_IS_NOW_LIMITED_TO_240_FPS, VSYNC_ENABLED, FPS_IS_NOW_CUSTOM_LIMITED_TO, HOME
+    }
+}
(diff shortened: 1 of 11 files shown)


Swan:
Hardcoding each translation isn't an option. I'll have to setup a submodule specifically for translations. We can do this in the form of language files which are setup similar to .ini where it's just `Key = Value`. Json wouldn't be the most optimal choice since it adds a lot of unnecessary bytes for what we're trying to accomplish.

Here's how I'd expect it to work:

1. You have *one* enum containing each possible key.
2. You load in the default file (English) and custom language file, and parse each key and value to a single dictionary.
3. Use `LocalizationManager.Get()` to get the appropriate string.

The actual implementation of localization should be held off for now since we plan to overhaul the entire UI, so that part can be removed. We're looking for just functionality right now, so that we can begin using it when we starting working on the new UI.

Quaver.Shared/Localization/LocalizedString.cs
```diff
@@ -0,0 +1,14 @@
...
+using System.Text;
+
+namespace Quaver.Shared.Localization
+{
+    /// <summary>
+    ///     Parameter to <see cref="LocalizationManager.Get(LocalizedString, object[])"/>
+    /// </summary>
+    public enum LocalizedString
+    {
+        FPS_IS_NOW_UNLIMITED, FPS_IS_NOW_LIMITED_TO_240_FPS, VSYNC_ENABLED, FPS_IS_NOW_CUSTOM_LIMITED_TO, HOME
```
Swan: This should be removed, but it's worth mentioning that each one should be on new lines and also not fully capitalized.

Sublimity:
I think CSV should be used instead of INI if this happens.

Swan:
Why? It's significantly more difficulty to read

PR #762, opened 2019-05-05
Number display fixes, ComboPosX, rating display

### LINKED ISSUE #708: AccuracyDisplayPosY is ignored
Broken here: https://github.com/Quaver/Quaver/commit/4d55efe2c336a947e65045a5a3265a96dfeeb7f3#diff-1d85b151922e4db9db11b9f854decd5dL286

YaLTeR:
See commits. Fixes #708, the weird digit sizes in 100.00% acc and some other stuff.

I set the KPS counter and the new rating display to be a set default offset from the top rather than based on accuracy Y position height because:
- if you want to move your acc to be, say, somewhere in the middle of the playfield you probably want to leave the KPS counter where it was in the top right corner rather than it following the acc's Y position
- if variable height digits are used the KPS counter would jump around

This also breaks existing skins a tiny bit, namely:
- the combo pos Y might need to be adjusted to be slightly lower as it's now computed from the vertical center of the combo display rather than from idk where
- the KPS counter Y might need to be adjusted


YaLTeR:
Maybe I should simplify the thing in `NumberDisplay` by merging `recomputeWidth` and `recomputeHeight` into one since it won't introduce any performance hit. Could also get rid of it altogether and recompute every time, but there's a case of the score display which on average updates every frame and with monospace digits doesn't trigger any recomputes. But then with non-monospace digits it'd trigger both recomputes every frame anyway, so might be worth simplifying it anyway.

Swan:
I think it's mergeable for now since it's pretty easy to see what's going on there, but I wouldn't mind it being a bit simpler, and you can open up another PR w/ those changes.

PR #779, opened 2019-05-16
Keysounds and sound effects

YaLTeR:
Poggers

Example mapsets: [7k fully keysounded pack](https://osu.ppy.sh/beatmapsets/577617), [o2jam converts with sound effects audio and keysounds](https://osu.ppy.sh/beatmapsets/827408#mania/1733808), [4k ranked sound effects and keysounds](https://osu.ppy.sh/beatmapsets/302726#mania/678460).

Videos: [7k keysounded pack](https://streamable.com/o5jn9), [that 4k map](https://streamable.com/nbezj). Sounds weird because I have big sample delay which I still haven't figured out.

`CustomAudioSampleCache` should probably be its own static class or something and not inside `GameplayScreen`, cc @Swan.

As far as I can tell everything works.

Requires https://github.com/Quaver/Wobble/pull/65 and https://github.com/Quaver/Quaver.API/pull/63.

### DIFF
--- a/Quaver.Shared/Screens/Editor/EditorScreen.cs
+++ b/Quaver.Shared/Screens/Editor/EditorScreen.cs
@@ -627,8 +627,13 @@ namespace Quaver.Shared.Screens.Editor
                 {
                     var view = (EditorScreenView) View;
 
-                    if (ConfigManager.EditorEnableHitsounds.Value && !view.LayerCompositor.ScrollContainer.AvailableItems[obj.EditorLayer].Hidden)
-                        HitObjectManager.PlayObjectHitSounds(obj);
+                    if (!view.LayerCompositor.ScrollContainer.AvailableItems[obj.EditorLayer].Hidden)
+                    {
+                        if (ConfigManager.EditorEnableHitsounds.Value)
+                            HitObjectManager.PlayObjectHitSounds(obj);
+                        if (ConfigManager.EditorEnableKeysounds.Value)
+                            HitObjectManager.PlayObjectKeySounds(obj);
+                    }
 
                     HitSoundObjectIndex = i + 1;
                 }
--- a/Quaver.Shared/Screens/Editor/EditorScreen.cs
+++ b/Quaver.Shared/Screens/Editor/EditorScreen.cs
@@ -48,6 +48,7 @@ using Quaver.Shared.Screens.Gameplay.Rulesets.HitObjects;
 using Quaver.Shared.Screens.Menu;
 using Quaver.Shared.Screens.Select;
 using Wobble;
+using Wobble.Audio.Tracks;
 using Wobble.Bindables;
 using Wobble.Graphics;
 using Wobble.Graphics.UI.Dialogs;
@@ -410,17 +411,17 @@ namespace Quaver.Shared.Screens.Editor
         {
             try
             {
-                if (AudioEngine.Track != null && AudioEngine.Track.IsPaused && !AudioEngine.Track.IsPreview)
+                if (AudioEngine.Track != null && AudioEngine.Track.IsPaused)
                     return true;
 
-                AudioEngine.LoadCurrentTrack();
+                AudioEngine.LoadCurrentTrack(false, WorkingMap.Length + 60000);
                 return true;
             }
             catch (Exception e)
             {
                 NotificationManager.Show(NotificationLevel.Error, "Audio track was unable to be loaded for this map.");
                 Exit(() => new MenuScreen());
-                return false;
+                return true;
             }
         }
 
@@ -715,7 +716,10 @@ namespace Quaver.Shared.Screens.Editor
                 if (AudioEngine.Track != null)
                     AudioEngine.Track.Rate = 1.0f;
 
-                AudioEngine.Track?.Fade(0, 100);
+                var track = AudioEngine.Track;
+
+                if (track is AudioTrack t)
+                    t?.Fade(0, 100);
 
                 return new SelectScreen();
             });
--- a/Quaver.Shared/Screens/Editor/EditorScreen.cs
+++ b/Quaver.Shared/Screens/Editor/EditorScreen.cs
@@ -173,6 +173,8 @@ namespace Quaver.Shared.Screens.Editor
             if (!LoadAudioTrack())
                 return;
 
+            GameplayScreen.CustomAudioSampleCache.LoadSamples(MapManager.Selected.Value, MapManager.Selected.Value.Md5Checksum);
+
             SetHitSoundObjectIndex();
 
             GameBase.Game.IsMouseVisible = true;
--- a/Quaver.Shared/Screens/Editor/EditorScreen.cs
+++ b/Quaver.Shared/Screens/Editor/EditorScreen.cs
@@ -173,7 +173,7 @@ namespace Quaver.Shared.Screens.Editor
             if (!LoadAudioTrack())
                 return;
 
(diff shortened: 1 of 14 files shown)


YaLTeR:
One thing I noticed is that the MD5 I use doesn't update with the file changes, so if I change a .qua on disk, then play it again, the custom audio samples won't be reloaded. Not sure how much of an issue that is and if it's an issue in the editor.

Quaver.Shared/Screens/Gameplay/CustomAudioSampleCache.cs
```diff
@@ -0,0 +1,104 @@
...
+ * Copyright (c) Swan & The Quaver Team <support@quavergame.com>.
+*/
+
+using System.Collections.Generic;
+using Quaver.Shared.Database.Maps;
+using Wobble.Audio.Samples;
+
+namespace Quaver.Shared.Screens.Gameplay
+{
+    public class CustomAudioSampleCache
```
Swan: Should be a static class.

Quaver.Shared/Screens/Gameplay/GameplayScreen.cs
```diff
@@ -225,6 +225,16 @@ public class GameplayScreen : QuaverScreen
         /// </summary>
         public Metronome Metronome { get; }
 
+        /// <summary>
+        ///     Loads and caches the custom audio samples.
+        /// </summary>
+        static CustomAudioSampleCache CustomAudioSampleCache { get; } = new CustomAudioSampleCache();
```
Swan: Remove this. Class should be static as mentioned

Quaver.Shared/Screens/Gameplay/CustomAudioSampleCache.cs
```diff
@@ -0,0 +1,104 @@
...
+        ///     Loads audio samples for the specified map into the cache.
+        /// </summary>
+        /// <param name="map"></param>
+        public void LoadSamples(Map map)
+        {
+            // Always clean up the left-over channels.
+            StopAll();
+
+            // If the MD5 is the same, no need to re-load the samples.
+            if (MapMd5 == map.Md5Checksum)
```
Swan: `map.Md5Checksum` will use the md5 that is cached in the database. Use `GameplayScreen.MapHash` if you want to retrieve the most up-to-date one.

Quaver.Shared/Screens/Gameplay/GameplayScreen.cs
```diff
@@ -260,6 +270,15 @@ public class GameplayScreen : QuaverScreen
 
             Timing = new GameplayAudioTiming(this);
 
+            // Initialize the custom audio sample cache and the sound effect index.
+            if (!IsCalibratingOffset)
+                CustomAudioSampleCache.LoadSamples(MapManager.Selected.Value);
+
+            NextSoundEffectIndex = 0;
```
Swan: I think it would be worth putting this into its own method.

Quaver.Shared/Screens/Gameplay/GameplayScreen.cs
```diff
@@ -815,6 +844,12 @@ private void SkipToNextObject()
                     var inputManager = (KeysInputManager)Ruleset.InputManager;
                     inputManager.ReplayInputManager.HandleSkip();
                 }
+
+                // Stop all playing sound effects and move NextSoundEffectIndex ahead.
+                CustomAudioSampleCache.StopAll();
+                while (NextSoundEffectIndex < Map.SoundEffects.Count &&
```
Swan: Code is duplicated here. As mentioned, you can make this a method and call it here as well.

Quaver.Shared/Screens/Editor/EditorScreen.cs
```diff
@@ -625,7 +625,7 @@ private void PlayHitsounds()
                 {
                     var view = (EditorScreenView) View;
 
-                    if (ConfigManager.EditorEnableHitsounds.Value && !view.LayerCompositor.ScrollContainer.AvailableItems[obj.EditorLayer].Hidden)
```
Swan: Reason for removing this? Doing so makes hitsounds played in the editor even if the user doesn't want to.

Quaver.Shared/Screens/Gameplay/CustomAudioSampleCache.cs
```diff
@@ -26,6 +26,11 @@ public class CustomAudioSampleCache
         /// </summary>
         private List<AudioSample> Samples { get; set; } = new List<AudioSample>();
 
+        /// <summary>
+        ///     Whether the corresponding audio samples are unaffected by rate.
+        /// </summary>
+        private List<bool> UnaffectedByRate { get; set; } = new List<bool>();
```
Swan: Instead of keeping two lists, it would be better to have one list with a class that looks as follows

```cs
public class GameplayAudioSample : IPlayableAudio, IDisposable
{
     public AudioSample Sample { get; }
     public bool UnaffectedByRate { get; }

     public GameplayAudioSample(...)
     {
     }

     public void Dispose() => Sample?.Dispose();
}
```

Swan:
It's also worth mentioning that fully keysounded charts such as the ones in the 7k pack can't be loaded in the editor and crashes the game.

Quaver.Shared/Screens/Gameplay/CustomAudioSampleCache.cs
YaLTeR: Where? Same folder it's currently in is fine?

YaLTeR:
> It's also worth mentioning that fully keysounded charts such as the ones in the 7k pack can't be loaded in the editor and crashes the game.

Yeah, you also can't change rates when one of them is selected. It's an unrelated issue I believe.

Quaver.Shared/Screens/Gameplay/CustomAudioSampleCache.cs
Swan: Quaver.Audio would be more fitting

Swan:
Pretty sure it just has to do with the fact that there is no valid audio file and that's what the editor/rate changes heavily rely on rather than some underlying clock.

Quaver.Shared/Screens/Gameplay/CustomAudioSampleCache.cs
YaLTeR: What about when it's called from `EditorScreen`?

Quaver.Shared/Screens/Editor/EditorScreen.cs
YaLTeR: Oh, I completely overlooked the fact that it was a different config option for the editor.

YaLTeR: Since I think keysounds and hitsounds should be separate settings, should there be a separate `PlayObjectKeySounds` method then?

Quaver.Shared/Screens/Gameplay/CustomAudioSampleCache.cs
Swan: You can use the cached version. Doesn't truly matter in this case since it gets updated upon saving.

Quaver.Shared/Screens/Editor/EditorScreen.cs
Swan: Yes

YaLTeR:
All done. Hopefully I didn't break anything while rebasing.

PR #792, opened 2019-05-20
Sound fixes

### LINKED ISSUE #420: Hitsound audio delay
**Describe the bug**
Hitsounds have audio delay and it's variable.

**To Reproduce**
1. Use the default sound-hit.wav.
2. Open some map, tap the keys loudly so you can hear both the hitsound and the key sound, before the music starts.
3. Quick-retry the map and repeat the experiment.

Expected: the hitsounds happen audibly at the same time as the physical key sounds.
Actual: the hitsounds are sometimes delayed, and sometimes happen at the same time. Usually at the very start they are more delayed than about a second later. Sometimes the hitsounds are really delayed compared to the physical key sounds.

Arch Linux x64, Quaver from #403

### LINKED ISSUE #511: Effect Volume changing when it should not change
**Description**
When changing Effect Volume or Master Volume, Effect Volume will change its value to whichever of these two was changed last (As seen in the videos below).

**Expected Behaviour**
Effect Volume should scale with Master Volume and itself (for example: Master Volume = 50%, Effect Volume = 0%, 50% * 0% = 0% = Actual Effect Volume) instead of just overwriting itself whenever either of those is changed.

**Additional Information**
https://streamable.com/ylhgk
https://streamable.com/htfl3

### LINKED ISSUE #638: Song volume isn't affected during transitions

YaLTeR:
Requires https://github.com/Quaver/Wobble/pull/67. Fixes #420, fixes #511, fixes #638 as far as I can tell. I'll need to write a wiki page or something on how to adjust the audio device properties on Linux properly to achieve low latency.

### DIFF
--- a/Quaver.Shared/Screens/Settings/SettingsDialog.cs
+++ b/Quaver.Shared/Screens/Settings/SettingsDialog.cs
@@ -8,12 +8,15 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading.Tasks;
+using ManagedBass;
 using Microsoft.Xna.Framework;
 using Microsoft.Xna.Framework.Input;
 using Quaver.Shared.Assets;
 using Quaver.Shared.Config;
 using Quaver.Shared.Graphics;
 using Quaver.Shared.Graphics.Backgrounds;
+using Quaver.Shared.Graphics.Dialogs;
 using Quaver.Shared.Graphics.Notifications;
 using Quaver.Shared.Graphics.Transitions;
 using Quaver.Shared.Helpers;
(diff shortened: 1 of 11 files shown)


YaLTeR:
:white_check_mark: perfect hitsound latency on my desktop
:white_check_mark: perfect hitsound latency on my laptop (-60ms to -150ms osu latency btw)
:white_check_mark: perfect hitsound latency for @Sublimity

Sublimity:
so this is why i got an email notification xd

Quaver.Shared/Screens/Settings/SettingsDialog.cs
```diff
@@ -309,7 +309,26 @@ private void CreateSections()
                     new SettingsSlider(this, "Effect Volume", ConfigManager.VolumeEffect, x => $"{x}%"),
                     new SettingsBool(this, "Pitch Audio With Rate", ConfigManager.Pitched),
                     new SettingsSlider(this, "Global Audio Offset", ConfigManager.GlobalAudioOffset, x => $"{x} ms"),
-                    new SettingsCalibrateOffset(this, "Calibrate Offset")
+                    new SettingsCalibrateOffset(this, "Calibrate Offset"),
+                    new SettingsSlider(this, "Audio Device Period", ConfigManager.DevicePeriod, x => $"{x} ms"),
+                    ((Func<Drawable>)(() =>
```
Swan: Instead of this, I would suggest creating a property for `Audio Device Period` and overriding the `Audio Device Buffer Length` slider to take in the device period slider as an argument. Then you can hook onto its bindable like normal. If you do it this way, OnDestroyed won't be needed, and you can unsubscribe from the bindable directly in the slider's `Destroy()` method. 

If you have to expose a few properties in the slider and make them public to make it happen, then I wouldn't mind it, but we should try to not do things inline like this for the sake of keeping the codebase clean

YaLTeR:
Done. I also added one more commit to fix a recent regression which prevented rates from working on virtual audio track maps.

Swan:
Awesome

PR #813, opened 2019-06-05
Refactor GameplayScreen.HandlePauseInput()

### LINKED ISSUE #806: Cursor is not hidden if the pause menu is exited via "Continue"
**Describe the bug**
It _is_ hidden if the pause menu is exited via the escape button, but not when using Continue.

**To Reproduce**
1. Play a map.
2. Press escape to pause.
3. Press Continue.
4. Cursor isn't hidden.

YaLTeR:
Refactor `HandlePauseInput()` to make it more readable. Obsoletes #655, additionally fixes #806. @Swan please comment on the TODO I left in the last commit, I'm marking this as a draft PR until then.

Currently the replay is always exited on tap. Would it make more sense to make it obey the tap to pause option too, so you have to hold pause to exit a replay if it's disabled?

### DIFF
--- a/Quaver.Shared/Screens/Gameplay/GameplayScreen.cs
+++ b/Quaver.Shared/Screens/Gameplay/GameplayScreen.cs
@@ -36,6 +36,7 @@ using Quaver.Shared.Screens.Gameplay.Rulesets.Input;
 using Quaver.Shared.Screens.Gameplay.Rulesets.Keys;
 using Quaver.Shared.Screens.Gameplay.Rulesets.Keys.Playfield;
 using Quaver.Shared.Screens.Gameplay.UI.Offset;
+using Quaver.Shared.Screens.Select;
 using Quaver.Shared.Skinning;
 using Wobble;
 using Wobble.Audio;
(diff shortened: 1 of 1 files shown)


Quaver.Shared/Screens/Gameplay/GameplayScreen.cs
```diff
@@ -523,6 +523,12 @@ private void HandlePauseInput(GameTime gameTime)
                 {
                     OffsetConfirmDialog.Exit(this);
                 }
+                // Exit the playing replay.
+                else if (InReplayMode)
+                {
+                    CustomAudioSampleCache.StopAll();
+                    Exit(() => new SelectScreen());
```
Swan: Might want to be careful here and deactivate any mods that the play initially had on it. Example of this would be the "Paused" mod which is a hidden mod that dictates if a user paused throughout the play.

Quaver.Shared/Screens/Gameplay/GameplayScreen.cs
```diff
@@ -493,6 +493,7 @@ private void HandlePauseInput(GameTime gameTime)
                     return;
 
                 // Properly fade in now.
+                // TODO: why is this here instead of wherever the restart key press is handled?
```
Swan: From what I can see it's handling the specific case of letting go of the pause key before it fully fades in. I suppose you could move it, but it might have an unintended result because the restart and pause are treated as separate actions that control the transition. Up to you if you want to refactor this further.

YaLTeR:
I decided not to worry about that fade for this PR. It should probably be unified with the replay somehow anyway.

I made it so all mods are removed upon exiting a replay. In future this will need to be
- made consistent with what happens if you watch replay to the end,
- perhaps instead of clearing the mods we need to restore the mods the player had active before playing the replay.

YaLTeR:
I decided not to worry about that fade for this PR. It should probably be unified with the replay somehow anyway.

I made it so all mods are removed upon exiting a replay. In future this will need to be
- made consistent with what happens if you watch replay to the end,
- perhaps instead of clearing the mods we need to restore the mods the player had active before playing the replay.

PR #932, opened 2019-07-08
make quaver.sh actually a working file

rainyuu:


YaLTeR:
Sorry to inform you, but it _was_ an actually working file, just that nobody ever fixed the repo settings to treat it as LF always, so @Swan (who is making the Steam builds) is getting it with CRLF line endings which obviously breaks everything.

Please pull request that change instead and don't change the script (the changes you made will actually result in breakage with e.g. folder names with spaces if I'm not mistaken).

Please also verify that the original script works if line endings are replaced with LF (for example, using `dos2unix quaver.sh`). I'll keep this PR open until that's verified (or if you want to replace the change with the one I suggested above).

rainyuu:
it indeed does work after dos2unix, so i guess the packaging person has to be notified to that

PR #947, opened 2019-07-12
Made notifications better

Sublimity:
- Fixed the problem with notification text overhanging.
- Notification height now adjusts relative to text height.
- Fixed avatar/image alpha when fading out.
- Changed the `BorderColor` to the main accent colour.
- The texture for the icon will now scale properly and doesn't require extra space in the file to accommodate

Requires #946 for new DarkGray colour.

### :thonking_face: might add a notification manager dialog soon:tm:

### DIFF
--- a/Quaver.Shared/Graphics/Notifications/Notification.cs
+++ b/Quaver.Shared/Graphics/Notifications/Notification.cs
@@ -152,7 +152,7 @@ namespace Quaver.Shared.Graphics.Notifications
 
         private void AdjustHeightIfRequired()
         {
-            if (!(Content.Height > Height))
+            if (Content.Height < Height)
                 return;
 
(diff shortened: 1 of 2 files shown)


Quaver.Shared/Graphics/Notifications/Notification.cs
```diff
@@ -133,5 +150,14 @@ public override void Draw(GameTime gameTime)
 
             base.Draw(gameTime);
         }
+
+        private void AdjustHeightIfRequired()
+        {
+            if (!(Content.Height > Height))
```
Swan: `if (Content.Height < Height)`

Sublimity: :joy:

Issue #949, opened 2019-07-13
Error indicator in graph in result screen


Janko5:
Error indicator - two values that means the average value of too late hits (0 to +∞ (blue gradient)) and average value of too early hits (-∞ to 0 (orange gradient)). 
![graph](http://puu.sh/DS8dO/c8f0d37877.png)

The same thing exist in osu!
![error](https://i.ppy.sh/b55e4030216b1201e0ade04ec8c11eab03fa1ff4/68747470733a2f2f6f73752e7070792e73682f68656c702f77696b692f41636375726163792f696d672f706572666f726d616e63655f67726170682e6a7067)

I think this indicator would be very useful for solving problems with offset/hitposition/scroll because it would give the exact values ​​that we should take into account.

For example:
Error indicator on one map is -23ms and +30ms. This means that the average value of too late hits is 7ms greater than the average value of too early hits. This means that to click more accurately, we need to set the offset to -3.5ms so that both values ​​was aligned, or set a bit smaller scroll, or set the hitposition a bit lower

PR #952, opened 2019-07-14
Results screen additions and 1 ms local offset adjustment with Ctrl

YaLTeR:
image

Things to consider:
- I would guess that the main button to be used here is the local offset adjustment one, so it might make sense to emphasize it somehow to make it easier to click. Perhaps put the buttons horizontally, below the divider line?
- You want to click only one of these buttons once, and only when your global or local offset equals to that which you used when playing the score. Perhaps we should store the global and local offset used in the score and then base the new offset calculation off of that instead of the current offset? This way it will always set the offset to the correct value.
- Not sure what the best place would be to put `FilterHitStats()`. I think there needs to be some place in `ResultsScreen` where all this data is computed once for all sub-components to use it?
- Maybe "average" sounds better than "mean"?
- "Adjust Map Offset" instead of "Adjust Local Offset"?
- Thinking about it, the LN release judgement window scale should probably _not_ be applied when computing mean and stddev.

Fixes ...uhh, I was sure there were at least 2 issues about this stuff, but I guess not. I believe this offers a better solution than #949 though.

### DIFF
--- a/Quaver.Shared/Screens/Result/UI/ResultScoreContainer.cs
+++ b/Quaver.Shared/Screens/Result/UI/ResultScoreContainer.cs
@@ -11,6 +11,8 @@ using System.Linq;
 using Microsoft.Xna.Framework;
 using Microsoft.Xna.Framework.Graphics;
 using Quaver.API.Maps.Processors.Rating;
+using Quaver.API.Maps.Processors.Scoring;
+using Quaver.API.Maps.Processors.Scoring.Data;
 using Quaver.Shared.Assets;
 using Quaver.Shared.Database.Maps;
 using Quaver.Shared.Helpers;
@@ -90,6 +92,19 @@ namespace Quaver.Shared.Screens.Result.UI
         /// </summary>
         private ResultHitDifferenceGraph HitDifferenceGraphRaw { get; }
 
+        /// <summary>
+        ///     A ScoreProcessor which is more likely to be filled with hit stats than the one in
+        ///     ResultScreen. For example, this one will have stats loaded from a replay.
+        ///
+        ///     TODO: this should really be the ResultScreen processor.
+        /// </summary>
+        private ScoreProcessor Processor { get; }
+
+        /// <summary>
+        ///     Hit statistics computed for the current score.
+        /// </summary>
+        private HitStatistics HitStatistics { get; }
+
         /// <inheritdoc />
         /// <summary>
         /// </summary>
@@ -99,6 +114,12 @@ namespace Quaver.Shared.Screens.Result.UI
             Size = new ScalableVector2(WindowManager.Width - 56, 490);
             Image = UserInterface.ResultScorePanel;
             DestroyIfParentIsNull = false;
+            Processor = Screen.GetScoreProcessor();
+
+            if (Processor.Stats != null)
+                HitStatistics = Processor.GetHitStatistics();
+            else
+                HitStatistics = new HitStatistics();
 
             CreateTopHorizontalDividerLine();
             CreateHeaderBackground();
--- a/Quaver.Shared/Screens/Result/UI/ResultScoreContainer.cs
+++ b/Quaver.Shared/Screens/Result/UI/ResultScoreContainer.cs
@@ -286,7 +286,7 @@ namespace Quaver.Shared.Screens.Result.UI
             {
                 Parent = this,
                 Y = firstItem.Y + firstItem.TextValue.Y + firstItem.TextValue.Height + 15,
-                Size = new ScalableVector2(VerticalDividerLine.X, 1),
+                Size = new ScalableVector2(Width, 1),
                 Alpha = 0.45f
             };
         }
--- a/Quaver.Shared/Screens/Result/UI/ResultScoreContainer.cs
+++ b/Quaver.Shared/Screens/Result/UI/ResultScoreContainer.cs
@@ -10,16 +10,22 @@ using System.Collections.Generic;
 using System.Linq;
 using Microsoft.Xna.Framework;
 using Microsoft.Xna.Framework.Graphics;
+using Quaver.API.Helpers;
 using Quaver.API.Maps.Processors.Rating;
 using Quaver.API.Maps.Processors.Scoring;
 using Quaver.API.Maps.Processors.Scoring.Data;
 using Quaver.Shared.Assets;
+using Quaver.Shared.Config;
 using Quaver.Shared.Database.Maps;
+using Quaver.Shared.Graphics;
+using Quaver.Shared.Graphics.Dialogs;
+using Quaver.Shared.Graphics.Notifications;
 using Quaver.Shared.Helpers;
+using Quaver.Shared.Screens.Menu.UI.Navigation.User;
 using Wobble;
-using Wobble.Assets;
 using Wobble.Graphics;
 using Wobble.Graphics.Sprites;
+using Wobble.Graphics.UI.Dialogs;
 using Wobble.Window;
 
 namespace Quaver.Shared.Screens.Result.UI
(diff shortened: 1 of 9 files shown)


YaLTeR:
Todo offset adjustment should be mean divided by rate not just mean

Swan:
* >Perhaps put the buttons horizontally, below the divider line? - I would say place the buttons below the graph horizontally. You can make the buttons bigger. As a side note, we'll want to be using `SpriteTextBitmap` for text. `BorderedTextButton` uses the old SpriteText
* Sure
* Quaver.API possibly?
* Average
* It's called local offset everywhere else. No need to change it
* Why?

YaLTeR:
> * Why?

Imagine if autoplay is playing an LN map with a "wrong offset", so all hits will be, say, 10 ms off. Currently mean will show less than 10 ms because half of the hits are LN releases, so they will be scaled 1.5x down. However, what we really want here is to show exactly 10 ms and to adjust the offset by 10 ms.

YaLTeR:
Also could you comment on storing the offsets in the score (question 2)?

YaLTeR:
image

YaLTeR:
TODO: the stuff done in `GetScoreProcessor()` should be done once and cached.

Quaver.Shared/Screens/Result/UI/ResultScoreContainer.cs
```diff
@@ -69,6 +73,11 @@ public class ResultScoreContainer : Sprite
         /// </summary>
         public Sprite ResultKeyValueItemDividerLine { get; private set; }
 
+        /// <summary>
+        ///     The horizontal divider line under the statistics key/value items.
+        /// </summary>
+        public Sprite StatisticsKeyValueItemDividerLine { get; private set; }
```
Swan: Instead of having this, just extend `ResultKeyValueItemDividerLine` to the full width

Quaver.Shared/Screens/Result/UI/ResultScoreContainer.cs
```diff
@@ -270,6 +280,125 @@ private void CreateKeyValueItems()
...
 
+        /// <summary>
+        ///     Creates all of the statistics items.
+        /// </summary>
+        private void CreateStatisticsKeyValueItems()
+        {
+            var processor = Screen.GetScoreProcessor();
+            var largestHitWindow = processor.JudgementWindow.Values.Max();
+
+            double mean = 0.0, standardDeviation = 0.0;
```
Swan:
```cs
double mean;
double standardDeviation;
```

Quaver.Shared/Screens/Result/UI/ResultScoreContainer.cs
```diff
@@ -270,6 +280,125 @@ private void CreateKeyValueItems()
...
+                    // Since variance(ax) = a^2 variance(x), then std(ax) = a std(x)
+                    standardDeviation /= ModHelper.GetRateFromMods(processor.Mods);
+                }
+            }
+
+            var availableWidth = Width - VerticalDividerLine.X;
+            var buttonPadding = 15;
+            var buttonWidth = (availableWidth - buttonPadding * 3) / 2;
+
+            var localOffsetButton = new BorderedTextButton("Adjust Local Offset", Colors.MainAccent,
```
Swan: You shouldn't be creating buttons here if the method is supposed to be creating KeyValue items

Swan: `BorderedTextButton` uses the older text. It's obsolete and should never be used. You can either create a new button or modify `TextButton` to use `SpriteTextBitmap`

Quaver.Shared/Screens/Result/UI/ResultScoreContainer.cs
```diff
@@ -270,6 +280,125 @@ private void CreateKeyValueItems()
...
+                (o, e) =>
+                {
+                    // Local offset is scaled with rate, so the adjustment depends on the rate the
+                    // score was played on.
+                    var change = mean * ModHelper.GetRateFromMods(processor.Mods);
+                    ResultScreen.Map.LocalOffset =
+                        (int) Math.Round(ResultScreen.Map.LocalOffset - change);
```
Swan: Put on the same line

Swan: If `change` isn't being used anywhere else, there's no need to make a local variable for it

Quaver.Shared/Screens/Result/UI/ResultScoreContainer.cs
```diff
@@ -270,6 +280,125 @@ private void CreateKeyValueItems()
...
+
+                    MapDatabaseCache.UpdateMap(ResultScreen.Map);
+                    NotificationManager.Show(NotificationLevel.Success,
+                        $"Local offset was set to {ResultScreen.Map.LocalOffset} ms.");
```
Swan: Put on the same line (and everything else similar to this)

Quaver.Shared/Screens/Result/UI/ResultScoreContainer.cs
```diff
@@ -270,6 +280,125 @@ private void CreateKeyValueItems()
...
+                X = VerticalDividerLine.X + buttonPadding,
+                Y = BottomHorizontalDividerLine.Y - 15 - 25,
+                Height = 25,
+                Width = buttonWidth,
+                Text =
+                {
+                    FontSize = 9
+                }
+            };
+            var globalOffsetButton = new BorderedTextButton("Adjust Global Offset", Colors.MainAccent,
```
Swan: The one issue I have with these buttons are that there should probably be some sort of confirmation button making the user aware of what their new offset will be and if they'd like to proceed. It can be in the form of a `ConfirmCancelDialog`. If not this, then the option to undo should be there, but I prefer the former.

Swan: Also, it's probably only worth creating these buttons only after gameplay. You wouldn't want to always display it (especially when clicking on an online leaderboard score or watching a replay)

Quaver.Shared/Screens/Result/UI/ResultScoreContainer.cs
YaLTeR: The local variable makes the expression more readable.

Quaver.Shared/Screens/Result/UI/ResultScoreContainer.cs
YaLTeR: There is value in having them on local scores: I used them a couple of times to adjust the offset from a score that I had before this was implemented, and also a couple of times when I forgot to click the button before escaping out of the gameplay score.

Quaver.Shared/Screens/Result/UI/ResultScoreContainer.cs
YaLTeR: > `BorderedTextButton` uses the older text. It's obsolete and should never be used. You can either create a new button or modify `TextButton` to use `SpriteTextBitmap`

Sounds out of scope for this PR. Bordered text buttons are used in a bunch of places still, so a subsequent PR could change them to use the new text.

YaLTeR:
Rebased and addressed all comments. Now depends on https://github.com/Quaver/Quaver.API/pull/67.

I'd really prefer _not_ to do the `ScoreProcessor` refactor outlined in the TODO in this PR.

Quaver.Shared/Screens/Result/UI/ResultScoreContainer.cs
Swan: There are some instances where you don't have access to these stats such as online scores and multiplayer (for other players). Hiding it when `Score.IsOnline` is true should do the trick.

Swan:
@YaLTeR  Quaver.API submodule can also be updated

PR #978, opened 2019-08-07
Add Visual Test Runner (Hotloading) & Unit Test Project

Swan:
- Added a project for future unit tests
- Setup hotloading for visual tests. It can be used for testing certain ui elements and/or quickly building screens without having to constantly restart the game. [Here it is in action](https://streamable.com/km7zm). To use it, compile using the **Visual Tests** solution configuration, and edit files as you normally would.


Quaver/Program.cs
```diff
@@ -111,7 +112,11 @@ private static void Run()
                 Logger.Error(e, LogType.Runtime);
             }
 
+#if VISUAL_TESTS
+            using (var game = new QuaverGame(new HotLoader("../../../../Quaver.Shared/")))
```
YaLTeR: Is there a better way? This will break with testing published builds.

Swan: How? And why would you test with a published build in the first place?

YaLTeR: > How?

Maybe there's some project variable with source location or something? Should at least make it overridable via an environment variable.

> And why would you test with a published build in the first place?

There already are some things that for some unknown reason work only in the published build (IPC). So I guess that's my reasoning. Even though you probably won't need IPC when building a UI through the visual tests.

PR #1174, opened 2019-10-22
Fix skin and scroll speed scale

### LINKED ISSUE #1136: Add HealthBarScale
Similar to OtherThingsScale.

### LINKED ISSUE #1165: [redesign] Overall scaling is off, especially in gameplay
Not sure if it should be described as a new thing or a bug, but everything is smaller in comparison with current Quaver. Not that I have anything aginst that itself, I quite enjoy it actually. The only problem is that it messess with the gameplay/skin settings when using the same skin as in old Quaver. Especially the column widths.

YaLTeR:
Requires https://github.com/Quaver/Quaver.Resources/pull/10

Notably absent: all constant sizes and offsets, although it's not a big deal at all, and most will probably be changed anyway with the gameplay screen redesign. Fixes #1165, fixes #1136 

I don't particularly like scaling integers by 1.4. Maybe all these properties should become floats? Would make more sense since all of them are used as floats in the end.


Quaver.Shared/Skinning/SkinKeys.cs
```diff
@@ -299,9 +299,64 @@ internal SkinKeys(SkinStore store, GameMode mode)
...
+        /// </summary>
+        /// <param name="value"></param>
+        /// <returns></returns>
+        private static int Scale(int value) => value * 1920 / 1366;
+        private static byte Scale(byte value) => (byte) (value * 1920 / 1366);
+
+        /// <summary>
+        ///     Scale all necessary skin keys to match the UI redesign coordinate system.
+        /// </summary>
+        private void FixScale()
```
Swan: Already talked about this, but this manual scaling can/should be solved by using attributes & reflection.

Here's an example:

`ScalableSkinValue.cs`
```cs
public class ScalableSkinValue : System.Attribute
{
}
```

`SkinKeys.cs`

```cs
[ScalableSkinValue]
public int ColumnSize { get; private set; }

[ScalableSkinValue]
public int ReceptorPosOffsetY { get; private set; }
```

Then after loading both default and user config, you can use reflection to set the scaled values:

```cs
public SkinKeys(SkinStore store, GameMode mode)
{
    Store = store;
    Mode = mode;

    ReadConfig(true);
    ReadConfig(false);
    LoadTextures();

    var props = GetType().GetProperties()
                .Where(prop => Attribute.IsDefined(prop, typeof(ScalableSkinValue))).ToList();

    const int scale = 1920 / 1366;

    foreach (var prop in props)
    {
        var value = prop.GetValue(this);
        
        switch (value)
        {
            case int val:
                prop.SetValue(this, val * scale);
                break;
            case float val:
                prop.SetValue(this, val * scale);
                break;
            case byte val:
                prop.SetValue(this, val * scale);
                break;
        }
    }
}
```

Quaver.Shared/Screens/Gameplay/Rulesets/Keys/HitObjects/HitObjectManagerKeys.cs
```diff
@@ -38,7 +38,7 @@ public static float ScrollSpeed
             get
             {
                 var speed = MapManager.Selected.Value.Qua.Mode == GameMode.Keys4 ? ConfigManager.ScrollSpeed4K : ConfigManager.ScrollSpeed7K;
-                return speed.Value / (20f * AudioEngine.Track.Rate);
+                return speed.Value / (20f * AudioEngine.Track.Rate) * 1920 / 1366;
```
Swan: Would be good to have the scale somewhere global

YaLTeR: Yea, where would a good place be?

Swan: The `QuaverGame` class should be fine

PR #1208, opened 2019-11-01
Fix Notification Max Width

Swan:


Quaver.Shared/Graphics/Notifications/DrawableNotification.cs
```diff
@@ -185,12 +185,14 @@ private void CreateIcon()
...
+            const int padding = 14;
+
             Text = new SpriteTextPlus(FontManager.GetWobbleFont(Fonts.LatoBlack), "", 20)
             {
                 Parent = this,
                 Alignment = Alignment.MidLeft,
-                X = Icon.X + Icon.Width + 14,
-                MaxWidth = Width - Icon.Width - 32,
+                X = Icon.X + Icon.Width + padding,
+                MaxWidth = Width - Icon.Width - Icon.X - padding * 1.5f,
```
YaLTeR: Well, it would actually make more sense to have padding = 7, then padding * 2 in X and padding * 3 in here.

PR #1265, opened 2019-11-15
Fix Game Freeze via TaskHandler

staravia:
Apparently logging inside a Task of a TaskHandler froze the game since it uses a different thread to run those tasks.


staravia:
oops wrong branch

PR #1266, opened 2019-11-15
Fixed TaskHandler tasks bug

staravia:
Apparently logging inside a Task of a TaskHandler froze the game since it uses a different thread to run those tasks.


YaLTeR:
> Apparently logging inside a Task of a TaskHandler froze the game since it uses a different thread to run those tasks.

That... doesn't make much sense, actually. Since OnCompleted is ran on the same thread as the actual function, only inside an additional lock. Are you sure this part really is the issue, and not the other rearrangements you did?

Swan:
A detailed explanation of why the freezing was happening can be found at #1269. Closing this, as it doesn't fix the issue.

PR #1284, opened 2019-11-18
Playfield Tilt Config (Prototype)

staravia:
# What is this?
This new feature will allow players to play the game at a simulated perspective (similar to guitar hero and stepmania hallway/distant)

If you would like to check out a video, click here

## Config
A `PlayfieldTilt4K` and `PlayfieldTilt7K` Config value was created for this feature.

## Image Plane
To render the Playfield, I've created a 2d plane in the form of a trapezoid. The plane is a simple geometry that consists of 4096 quads. Because of how UV maps work, this much detail is needed. Illustration below:

## PreDraw()
A `PreDraw` method was also created in the Playfield and Ruleset abstracts due to xna wiping the renderer everytime we change Render Targets.

### DIFF
--- a/Quaver.Shared/Config/ConfigManager.cs
+++ b/Quaver.Shared/Config/ConfigManager.cs
@@ -174,6 +174,16 @@ namespace Quaver.Shared.Config
         internal static Bindable<ScrollDirection> ScrollDirection7K { get; private set; }
 
         /// <summary>
+        ///     Determines how much tilt is applied to the 4K Playfield.
+        /// </summary>
+        internal static BindableInt PlayfieldTilt4K { get; private set; }
+
+        /// <summary>
+        ///     Determines how much tilt is applied to the 7K Playfield.
+        /// </summary>
+        internal static BindableInt PlayfieldTilt7K { get; private set; }
+
+        /// <summary>
         ///     The offset of the notes compared to the song start.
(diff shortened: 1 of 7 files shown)


YaLTeR:
> PreDraw

You're looking for `ScheduledRenderTargetDraws` or something like that. Was added recently for that exact purpose.

YaLTeR:
> CreateOrthographicOffCenter

Wasn't orthographic, like, the opposite of perspective? Could that be the reason for the texture artifacts? Disclaimer: I don't know much about all this, but the texture artifacts seem extremely weird to me.

staravia:
> 
> 
> > CreateOrthographicOffCenter
> 
> Wasn't orthographic, like, the opposite of perspective? Could that be the reason for the texture artifacts? Disclaimer: I don't know much about all this, but the texture artifacts seem extremely weird to me.

Yes that is true, but the entire game is in orthographic view. While we can create a perspective camera, everything will have to realign to the original playfield because we want to interpolate between how much it tilts and the default.

YaLTeR:
Perspective transform is a matrix transform just like anything else. You should be able to render the playfield with a perspective transformation, and then render the rest in screen-space like it's done now.

Swan:
Gonna close since there hasn't been progress on this for a while.

Issue #1546, opened 2020-02-10
A better way to set hit positions and whatnot in skins


YaLTeR:
This is a proposal for changing how the hit positions are set in skins, which should fix the current issues related to it, namely some weirdness with setting it up, and the existence of the "center lines" checkbox in the editor.

The proposal is to change whatever the current hit position variables are to these two:
1. offset on the receptor image,
1. offset on the note image,

and then the note hit position is where these two offsets align (and set the timing line offset accordingly, etc.). This will get rid of the editor checkbox (it'll be handled by design), and will also allow to naturally handle receptors with gradients on top, and notes with non-trivial shapes.

Examples:
![image](https://user-images.githubusercontent.com/1794388/74159828-777ad780-4c2d-11ea-9d5d-89c07e6d5e27.png)
![image](https://user-images.githubusercontent.com/1794388/74159946-adb85700-4c2d-11ea-889a-a9062f5775de.png)
![image](https://user-images.githubusercontent.com/1794388/74160351-4fd83f00-4c2e-11ea-966c-ed86fd90c996.png)
![image](https://user-images.githubusercontent.com/1794388/74160745-ed337300-4c2e-11ea-8acb-ada7c8795890.png)

YaLTeR:
After thinking a bit more on this, I realized the note hit offsets and receptor hit offsets are not completely independent. For example, if you want to use arrow/circle-type receptors with bars, you'll want the bar hit offsets to be centered (and the timing lines to be aligned with the centers of the bars). Think circle skins but with squares instead of circles.

So maybe we should keep the current receptor offset (which would behave as described above), and add another option for "set note hit offset to center", basically the checkbox from the editor but put in skin.ini. I'm not sure there are any notes which need something other than bottom (bars) and center (arrows, circles, etc.)?

Issue #453, opened 2019-01-02
Replay System Improvements


YaLTeR:
Potential nice features for play and choke analysis:
- [X] Pausing, slowing down, speeding up, forwards and backwards seeking in replays. Think video player controls.
- Visualizing hit windows right on the playfield.
- [X] Visualizing key presses and releases (for LNs) on the playfield. Goes along with the previous point, makes it possible to visually see where your button press was and where it needed to be.
- SV toggle.
- Having two synchronized playfields side by side, one with SVs, another without SVs. Alternatively, somehow drawing notes without SV (like border / outline / wireframe) on top of the playfield. This would actually be great as a practice mod and not just in a replay viewer.

YaLTeR:
For inspiration on visualizing the key presses: https://twitter.com/transcendtoinf/status/1221346701283184640

PR #1547, opened 2020-02-10
Hit display for replays

YaLTeR:
cc #453

The hit positions are based on the timing line positions, which means that they look perfect on bars, but on (at least the default) arrows and circles they are aligned to the bottom of the hit objects, whereas you could argue that the expected would be to the center. This should really be solved by changing the timing line base and not by hardcoding something into the hit display (see #1546 for my idea on how to do that).

### DIFF
--- a/Quaver.Shared/Screens/Gameplay/Rulesets/Keys/Playfield/Hits/Hit.cs
+++ b/Quaver.Shared/Screens/Gameplay/Rulesets/Keys/Playfield/Hits/Hit.cs
@@ -1,0 +1,206 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ * Copyright (c) Swan & The Quaver Team <support@quavergame.com>.
+*/
+
+using System;
+using System.Collections.Generic;
+using Microsoft.Xna.Framework;
+using Quaver.API.Enums;
+using Quaver.API.Maps.Processors.Scoring.Data;
+using Quaver.Shared.Config;
+using Quaver.Shared.Screens.Gameplay.Rulesets.Keys.HitObjects;
+using Quaver.Shared.Skinning;
+using Wobble.Graphics;
+using Wobble.Graphics.Sprites;
+
+namespace Quaver.Shared.Screens.Gameplay.Rulesets.Keys.Playfield.Hits
+{
+    public class Hit
+    {
+        /// <summary>
+        ///     The hit stat for this hit.
+        /// </summary>
+        private HitStat? HitStat { get; set; }
+
+        /// <summary>
+        /// </summary>
+        private HitObjectManagerKeys Manager { get; }
+
+        /// <summary>
+        /// </summary>
+        private IDictionary<Judgement, Color> JudgeColors { get; }
+
+        /// <summary>
+        /// </summary>
+        private ScrollDirection ScrollDirection { get; }
+
+        /// <summary>
+        ///     Position for hitting without taking hit object height into account.
+        ///     Use this to compute positions of line indicators (height = 1), such as the press and release position.
+        /// </summary>
+        private float LineHitPosition { get; }
+
+        /// <summary>
+        ///     The indicator line itself.
+        /// </summary>
+        private Sprite Indicator { get; }
+
+        /// <summary>
+        ///     Vertical line from the indicator to the perfect position.
+        /// </summary>
+        private Sprite LineToPerfect { get; }
+
+        /// <summary>
+        ///     The actual hit position.
+        /// </summary>
+        private long Position { get; set; }
+
+        /// <summary>
+        ///     Perfect position for this hit.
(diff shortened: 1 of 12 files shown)


Quaver.Shared/Screens/Gameplay/Rulesets/Keys/Playfield/Hits/Hit.cs
```diff
@@ -0,0 +1,211 @@
...
+            var playfield = (GameplayPlayfieldKeys) ruleset.Playfield;
+            ScrollDirection = playfield.ScrollDirections[lane];
+            LineHitPosition = playfield.TimingLinePositionY[lane];
+
+            var laneX = playfield.Stage.Receptors[lane].X;
+            LineToPerfect = new Sprite
+            {
+                Alignment = Alignment.TopLeft,
+                Position = new ScalableVector2(laneX + 0.5f * playfield.LaneSize, 0),
+                Size = new ScalableVector2(1, 0),
```
Swan: It's typically not a good idea to use 1px size for lines right now, since the entire line can just disappear and cause a flashing effect. Use 2 instead.

Quaver.Shared/Screens/Gameplay/Rulesets/Keys/Playfield/Hits/HitManager.cs
```diff
@@ -0,0 +1,143 @@
+using System.Collections.Generic;
+using System.Diagnostics;
+using Quaver.API.Maps.Processors.Scoring.Data;
+using Quaver.Shared.Screens.Gameplay.Rulesets.Input;
+using Quaver.Shared.Screens.Gameplay.Rulesets.Keys.HitObjects;
+
+namespace Quaver.Shared.Screens.Gameplay.Rulesets.Keys.Playfield.Hits
+{
+    public class HitManager
```
Swan: Instead of creating an entirely new manager and container for this, why not have the hits be children of the hitobjects themselves? Creating a new pool for this seems a bit redundant when the hitobjects already have that system in place and the hits act that way anyway. It should be a lot simpler this way as well.

Quaver.Shared/Screens/Gameplay/Rulesets/Keys/GameplayRulesetKeys.cs
```diff
@@ -61,6 +69,20 @@ public bool ShowHits
...
                 _showHits = value;
 
+                var manager = (HitObjectManagerKeys) HitObjectManager;
+                foreach (GameplayHitObjectKeys hitObject in manager.ActiveNoteLanes.Concat(manager.DeadNoteLanes).Concat(manager.HeldLongNoteLanes).Flatten())
+                {
+                    var tint = hitObject.Tint * (_showHits ? 1 : ShowHitsNoteAlpha);
+                    var newTint = hitObject.Tint * (_showHits ? ShowHitsNoteAlpha : 1);
+
+                    hitObject.HitObjectSprite.Tint = tint;
+                    hitObject.HitObjectSprite.FadeToColor(newTint, Easing.OutQuad, 250);
```
Swan: You'll want to call `hitObject.[part_of_object].ClearAnimations()` before adding a new one, so you only have one fade animation running at a time.

Quaver.Shared/Screens/Gameplay/Rulesets/Keys/Playfield/Hits/Hit.cs
```diff
@@ -0,0 +1,211 @@
...
+        /// </summary>
+        private long PerfectPosition { get; set; }
+
+        /// <summary>
+        ///     Computes the hit position and the perfect hit position and returns them as a tuple.
+        /// </summary>
+        /// <param name="manager"></param>
+        /// <param name="hitStat"></param>
+        /// <returns>a tuple of (Position, PerfectPosition)</returns>
+        public static (long, long) ComputePositions(HitObjectManagerKeys manager, HitStat hitStat)
```
Swan: Create a struct/class for this instead of returning two values.

Quaver.Shared/Screens/Gameplay/UI/Replays/ReplayController.cs
```diff
@@ -370,6 +375,19 @@ private void CreateSpeedSlider()
...
         }
 
+        /// <summary>
+        /// </summary>
+        private void CreateShowHitsButton()
+        {
+            ShowHits = new ReplayControllerShowHits(Screen, new ScalableVector2(264, 38))
+            {
+                Parent = this,
+                Y = -14 - SpeedSlider.Height - 14,
```
Swan: Use a const for the -14

Quaver.Shared/Screens/Gameplay/Rulesets/Keys/Playfield/Hits/HitManager.cs
YaLTeR: That's what my initial implementation was, but the main issue was that there's no good correspondence between hit objects and their hit stats, so I ended up doing a linear search over all hit stats for every hit object, twice (press and release). It was actually like 2x more code (because you need to handle both presses and releases for each hit object).

PR #1677, opened 2020-03-10
Increase editor seeking speed while track is playing

### LINKED ISSUE #1674: Editor: Fast-scrolling w/ scroll wheel when song is playing
Look at how other games do it and you'll see what I mean. It is a horrible experience when the song is playing and you want to rewind a bit or go forward a bit.

Out of habit, some people like myself will intentionally play the song in the editor and use the scroll wheel to navigate because of how other games scroll faster when the song is playing.

Swan:
Closes #1674


staravia:
Thank  you!  💃  💃  💃

PR #1795, opened 2020-04-03
Implement new failure animation

### LINKED ISSUE #411: Gradual slowdown on fail
Currently when you fail the map stops scrolling instantly. This is a bit unpleasant to look at due to the motion misprediction from the brain.

Make the map slow down gradually, see osu!mania for example.

### LINKED ISSUE #869: Don't play fail animation / fail sound when quitting map
Retry doesn't have the red screen, fail sound and delay when pressing it, back to menu shouldn't have it either if "Skip Results Screen After Quitting" is toggled on. Every bit of navigation speed optimization helps so that you spend less time in menus and more time playing the game.

Swan:
Gets rid of that instant red screen and adds a gradual screen and track fade.

Only known issue is when playing rates with audio pitching off - it'll skip forward for a half second while its adjusting, so this should be fixed framework side.

Closes #411
Closes #869

### DIFF
--- a/Quaver.Shared/Screens/Gameplay/GameplayAudioTiming.cs
+++ b/Quaver.Shared/Screens/Gameplay/GameplayAudioTiming.cs
@@ -87,7 +87,7 @@ namespace Quaver.Shared.Screens.Gameplay
         public void Update(GameTime gameTime)
         {
             // Don't bother updating if the game is paused or the user failed.
-            if (Screen.IsPaused || Screen.Failed)
+            if (Screen.IsPaused)
                 return;
 
             if (Screen.IsMultiplayerGame && !Screen.IsMultiplayerGameStarted)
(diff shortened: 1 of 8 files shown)


YaLTeR:
Giving this a test. So far I noticed:
- the screen doesn't fade to red, it probably should
- the fail sound doesn't play until the end of the fade and it should play right away
- the timing lines still stop in place

YaLTeR:
Seems to work fine now. Just need the reset button to work during this for the perfect fail experience.

> Removed this on purpose. The red is a bit too much imo, but mostly because of how bright it is.

Oh yeah, that's definitely too bright. I was thinking something more subtle, like 10% red or something, and not fading to complete black (I kinda like what osu does here).

Swan:
Should be better now I think.

> Just need the reset button to work during this for the perfect fail experience.

This could be handled in another PR. An issue is already opened up for this

PR #1803, opened 2020-04-04
Temporary solution to force refresh mapsets.

### LINKED ISSUE #1185: [redesign] Game doesn't show/process new .qua files that's been placed into a mapset folder
In 0.17.0 & older versions, usually one would be able to see a new difficulty once a .qua is placed inside the folder of a mapset, but it doesn't in this case.
Makes it very difficult (if not, impossible) to add in new difficulties to your mapset manually.

Sublimity:
This temporary solution just allows for the end user to begin a force refresh if they so desire. This can be done with **F5** in the select screen, and **CTRL+F5** in the main menu, to flag for a force refresh when entering singleplayer.

The functions that are being re-used will have to be optimized at a later date. This is only to counter having to rebuild the cache on the old steam build/develop branch.

Closes #1185

### DIFF
--- a/Quaver.Shared/Screens/Main/MainMenuScreen.cs
+++ b/Quaver.Shared/Screens/Main/MainMenuScreen.cs
@@ -100,7 +100,7 @@ namespace Quaver.Shared.Screens.Main
                 return;
 
             HandleKeyPressEscape();
-            HandleKeyPressControlF5();
+            HandleKeyPressF5();
         }
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


YaLTeR:
> F5 in the select screen, and CTRL+F5 in the main menu

I don't like this difference in hotkeys.

Sublimity:
I figured I'd leave F5 free for something else in the main menu, it doesn't exactly make sense to use F5 in the menu, as there is nothing related to refreshing in that screen, and also because that button would only be used in times of need, such as if the user has no mapsets cached, and dragged some into their songs folder.

If you think it'd still be better to use F5 in that screen, I'll swap it over.

YaLTeR:
I think that there should not be a difference between the hotkeys for the same action on different screens.

Quaver.Shared/Screens/Main/MainMenuScreen.cs
```diff
@@ -107,6 +114,21 @@ private void HandleKeyPressEscape()
             DialogManager.Show(new QuitDialog());
         }
 
+        private void HandleKeyPressControlF5()
```
Swan: Not sure about doing it this way. If the user chooses to refresh here and clicks yes on the dialog, I would expect it to do it immediately rather than wait for them to go to song select.

PR #1857, opened 2020-04-09
Refactor base path detection to work if path does not contain "Etterna"

### LINKED ISSUE #1856: External game DB links do not work if the path contains a space
**Describe the bug**
When attempting to import an external game's .db file from a path that contains a space, Quaver will accept the .db file and write the correct path to the config file, but no songs will be loaded from it. Renaming the path to the db file to not contain any spaces fixes the problem.

**To Reproduce**
- Create a game directory (osu!/Etterna) that has a path containing a space. (e.g. `C:\Etterna with a space`)
- Launch Quaver and import that game's DB file. Observe that the correct path is written to `quaver.cfg`, but no songs are loaded.
- Exit Quaver.
- Rename the external game path so that it no longer contains a space. (e.g. `C:\Etterna`)
- Launch Quaver and re-import the DB file. Songs are now loaded (may require restarting the game again).

**Expected behavior**
The game should load files from the external DB.

**Screenshots**
I can do one better and include a video:
[Quaver DB path bug video.zip](https://github.com/Quaver/Quaver/files/4458754/Quaver.DB.path.bug.video.zip)

(If you don't feel comfortable downloading a random ZIP file on the internet, fair enough, [click here](https://cdn.discordapp.com/attachments/398037301696921600/697922923746426950/Quaver_DB_path_bug_-_re-encoded.mp4))

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser: Firefox
 - Version: Latest release

**Additional context**
As seen in the video, the game is still able to drop database lock files in the correct path. It just fails to read them

monotron:
This PR fixes #1856 - the load method was assuming the base Etterna path included the substring "Etterna" - if it does not, the path will be invalid.

### DIFF
--- a/Quaver.Shared/Database/Maps/Etterna/EtternaDatabaseCache.cs
+++ b/Quaver.Shared/Database/Maps/Etterna/EtternaDatabaseCache.cs
@@ -12,6 +12,8 @@ namespace Quaver.Shared.Database.Maps.Etterna
 {
     public class EtternaDatabaseCache
     {
+        private static readonly string cacheDatabaseRegularExpression = @"[\\/]Cache[\\/]cache\.db$";
+
         public static List<OtherGameMap> Load()
         {
             if (string.IsNullOrEmpty(ConfigManager.EtternaDbPath.Value))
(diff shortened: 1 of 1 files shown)


monotron:
It might be worthwhile testing this on Linux/macOS to make sure it works when the path separator is `/` instead of `\` like on Windows. I made sure the regex matches for a given path in regexr but we can never be too sure.

PR #1865, opened 2020-04-13
Added skin-able key click sfx for textbox's

Sublimity:
When a character is typed in the search filter on the selection screen, if there is at least 1 `sound-menu-keyclick-x.wav` sfx, then a randomized selection of those sound effects will be played.

[Preview](https://streamable.com/44xib4)

cc Quaver/Wobble#92
cc Quaver/Quaver.Wiki#70

### DIFF
--- a/Quaver.Shared/Skinning/SkinStore.cs
+++ b/Quaver.Shared/Skinning/SkinStore.cs
@@ -20,6 +20,7 @@ using Quaver.Shared.Config;
 using Wobble;
 using Wobble.Assets;
 using Wobble.Audio.Samples;
+using Wobble.Graphics.UI.Form;
 using Wobble.Logging;
 
 namespace Quaver.Shared.Skinning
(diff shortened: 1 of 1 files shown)


Quaver.Shared/Screens/Selection/UI/FilterPanel/Search/FilterPanelSearchBox.cs
```diff
@@ -121,5 +128,17 @@ private void StoppedTyping(string filter)
             CurrentSearchQuery.Value = filter;
             PreviousSearchTerm = filter;
         }
+
+        /// <summary>
+        ///		Called when the user types a character in the textbox.
+        ///	</summary>
+        private void CharacterTyped(char c)
```
Swan, line 135: Rather than adding this functionality for one single textbox, it should be implemented for all of them at framework level. Ideally when the skin loads, the list of AudioSamples should be cached for *every* textbox to use. Then each individual textbox can decide if they want to play the sounds or not (on by default).

Swan:
Also worth noting that a wiki PR should be included with this

Quaver.Shared/Screens/Selection/UI/FilterPanel/Search/FilterPanelSearchBox.cs
Sublimity: That sounds good actually, I didn't think of the other textboxes.

PR #1866, opened 2020-04-13
Add exiting OptionsDialog when clicked outside of main dialog

### LINKED ISSUE #1448: [redesign] clicking outside options doesn't close them

Sublimity:
Fixes #1448

### DIFF
--- a/Quaver.Shared/Screens/Options/OptionsDialog.cs
+++ b/Quaver.Shared/Screens/Options/OptionsDialog.cs
@@ -18,6 +18,12 @@ namespace Quaver.Shared.Screens.Options
         {
             FadeTo(0.75f, Easing.Linear, 200);
             CreateContent();
+
+            Clicked += (sender, args) =>
+            {
+                if (!Menu.IsHovered())
+                    Close();
+            };
         }
 
         /// <inheritdoc />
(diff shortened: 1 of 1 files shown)


Quaver.Shared/Screens/Options/OptionsDialog.cs
```diff
@@ -40,7 +40,8 @@ public override void CreateContent()
         /// <param name="gameTime"></param>
         public override void HandleInput(GameTime gameTime)
         {
-            if (KeyboardManager.IsUniqueKeyPress(Keys.Escape))
+            if (KeyboardManager.IsUniqueKeyPress(Keys.Escape) || MouseManager.IsUniqueClick(MouseButton.Left) &&
```
Swan, line 43: This needs to handle the case of when dropdowns extend out of the menu (resolution/skin).

Swan: Handling the `Clicked` event in the constructor should be enough which handle the depth of the buttons for you.

```cs
public OptionsDialog() : base(0)
{
    ...

    Clicked += (sender, args) =>
    {
        if (!Menu.IsHovered())
            Close();
    };
}
```

Issue #1690, opened 2020-03-12
Difficulty graph for map previews doesn't update when switching to 1.0x rate


Fretzi:
When switching from 1.1x or 0.95x to 1.0x, the graph next to the map preview does not update and adjust to the 1x speed.
This is most likely because 1.0x isn't actually a rate.

thenerdie:
This has just now been resolved.

Issue #1637, opened 2020-03-01
Inconsistency on scoreboard during gameplay for scores below 10qr


Fretzi:
While playing, previous scores show as single digit on the scoreboard (eg 5.14), while your current score shows as double digit (eg 05.12).
![2020-03-01_204608](https://user-images.githubusercontent.com/48097547/75632617-30657e00-5bfe-11ea-9a3f-6321006ebf1f.jpg)

thenerdie:
This is now fixed. I will now create a pull request.

PR #1904, opened 2020-04-23
Fix issues #1690, #1637

thenerdie:
I fixed 3 or so issues in the last 2 weeks and I didn't know I was supposed to create a new branch for each fix (oops).

Also added Green Number support in the options :D


Swan:
Same issue as your last opened PR. Needs one branch per fixed issue with proper commit messages.

Issue #1196, opened 2019-10-29
[redesign] Some NPS adjustments


robby250:
First of all I wanted to suggest that the NPS value should measure from the first note to the last note of the map, but it has already been mentioned here: https://github.com/Quaver/Quaver/issues/1190

Secondly, It appears that the NPS value is stored as an integer instead of a float, which causes all the maps with the same integer NPS (16 for example) to be sorted alphabetically instead of by NPS more precisely (i.e 16.12 NPS then 16.13 NPS etc). I believe that the NPS should be stored with at least 3-4 decimals for more accurate sorting and also for better search filtering (NPS>16.5 NPS=18.22 etc) and only display 1 or 2 decimals in the UI for aesthetic reasons.

PR #1927, opened 2020-05-04
Convert NPS value from int to float

robby250:
NPS will be more precise for sorting and viewing in the map banner. However for the banner I have rounded the value down to 1 decimal, because I don't think any more decimals are required (initially I chose 2 but I think 1 is enough).

closes https://github.com/Quaver/Quaver/issues/1196


robby250:
oops forgot to change the search part of NPS, it's still an int there

robby250:
NPS search seems to work with decimals now, and I have also changed the NPS mapset banner value from 1 decimal to 2.

robby250:
Decided to truncate the banner display value instead of rounding it, because it's probably better I think?

PR #2007, opened 2020-05-18
Fix local offset rate scaling

robby250:
Local offset didn't take into account rates.
On 80 ms local offset, this is a map on 1.0x with 10 scroll speed:
unknown

Same map on 0.5x, without any code changes:
unknown

Same map on 0.5x, after my proposed change:
unknown

While my rhythm sense is really bad, I can definitely feel the difference in offset when using the lower rate compared to 1.0x, but it should probably be tested by someone who doesn't suck at tapping through audio cues.
Also I probably hit later on the original 1.0x compared to the second 0.5x graph because it was significantly harder to play than 0.5x.


robby250:
nevermind i misunderstood how local offset works

PR #2033, opened 2020-05-23
Fix LN rendering with SV direction changes in the middle

### LINKED ISSUE #249: Implement Negative SV
Allow SV to be a negative value.

Precautions:
- Implement reverse direction LN
- Flip LN tail on reverse
- Might require changes on Quaver.API (when importing maps/validating maps)

YaLTeR:
[Video](https://cdn.discordapp.com/attachments/526869603515760659/713710244366123098/out.mp4)

LN bodies span from the earliest visible LN position to the latest visible LN position. LN ends are flipped if the LN ends during a negative SV.

Closes #249 at last.

I've tested this a bunch but something may come up as usual so I'll report after playing a session with this.

Unfortunately, this makes a "flashing" gimmick on https://quavergame.com/mapsets/map/11611 broken as it relied on the previous (broken) behavior of the LN rendering (already notified the author).

There's also a certain limitation of the current negative SV implementation that came to my mind: since negative SVs break note position non-decreasing, notes far in the future don't load in even if they should be currently on screen due to negative SVs later on in the map. I guess this isn't a big priority currently but it's something to keep in mind and perhaps make an issue about.

### DIFF
--- a/Quaver.Shared/Screens/Gameplay/Rulesets/Keys/HitObjects/GameplayHitObjectKeys.cs
+++ b/Quaver.Shared/Screens/Gameplay/Rulesets/Keys/HitObjects/GameplayHitObjectKeys.cs
@@ -161,7 +161,6 @@ namespace Quaver.Shared.Screens.Gameplay.Rulesets.Keys.HitObjects
             var playfield = (GameplayPlayfieldKeys)ruleset.Playfield;
             var posX = playfield.Stage.Receptors[lane].X;
             var flipNoteBody = direction.Equals(ScrollDirection.Up) && SkinManager.Skin.Keys[MapManager.Selected.Value.Mode].FlipNoteImagesOnUpscroll;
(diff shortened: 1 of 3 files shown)


YaLTeR:
I played for a little and everything was fine.

Swan:
> Unfortunately, this makes a "flashing" gimmick on https://quavergame.com/mapsets/map/11611 broken as it relied on the previous (broken) behavior of the LN rendering (already notified the author)

It's probably worth adding versioning to the .qua format now to keep compatibility given that this map is already ranked and also to future-proof any other gameplay mechanic changes.

A similar issue exists (scroll speed is extremely slow) with [another ranked map from 2018](https://quavergame.com/mapsets/106) from a different change which could have been avoided entirely if we had proper versioning in place.

YaLTeR:
I don't think it's worth for these particular changes since they affect a single ranked map each. While unfortunate, we are still in closed alpha after all and I think this gives us permission to do these two breaking changes.

It should certainly be done after we're out of closed alpha but I think it's better to not keep backwards compatibility hacks while we can afford that.

PR #2034, opened 2020-05-23
Fix map never finishing when it ends with a 0× SV

YaLTeR:
Based on #2033 because it has `CurrentVisualPosition`. Only 2 last commits are from this change.

wtf how is there no issue for this again, i guess i never made one


YaLTeR:
I brought the object count checks back because it occurred to me that on large judgement windows it might be possible to go beyond the 500 ms hardcoded in the previous version.

Now the check is similar to what it used to be, but if there are only dead LNs then it checks for visual time > map length.

Issue #1684, opened 2020-03-11
Don't rely on bass audio library times when drawing frames


robby250:
With bass: 
![76436712-39560c80-63c1-11ea-8136-ebf07b26bbe5](https://user-images.githubusercontent.com/51453670/76442255-5393e880-63c9-11ea-9164-e09f703771e1.jpg)

Without bass:
![20200311_184832](https://user-images.githubusercontent.com/51453670/76442166-33642980-63c9-11ea-9297-f2f36fc402d4.jpg)

Consistent frame distance, much smoother motion overall.

YaLTeR:
Obligatory note that audio and frame clocks are different so if audio timing isn't taken into account at all everything will go off-sync.

PR #2046, opened 2020-05-27
Change Time to use frame time instead of audio time

robby250:
Falling notes no longer rely on bass audio library time, as doing that causes constant stuttering/frameskips. Instead now it only syncs with bass time if the frame time falls behind, once every second (as a fallback in case the frame time runs ahead of audio time), and during the fail animation in order to keep the slowdown animation.

This should result in much smoother frames on high refresh rate monitors.

closes https://github.com/Quaver/Quaver/issues/1684


YaLTeR:
Don't change unconditionally and by default please.

robby250:
Added a toggleable option which is off by default to use frame time, otherwise it uses the old method.

robby250:
Moved the option from "Video" to "Miscellaneous".

Issue #944, opened 2019-07-11
Wayland support


YaLTeR:
Running natively on Wayland rather than through XWayland could allow properly supporting DPI scaling, potentially improve performance and make it possible to experiment with things like the presentation-time protocol for stutter prevention.

Currently it's possible to run Quaver on Wayland on Arch Linux by using the system libSDL2 via `LD_PRELOAD`. However:
- ~The window is partially transparent.~ Fixed by clearing the alpha channel after drawing.
- ~Mouse clicks aren't recognized. This is likely due to MonoGame [using `Sdl.Mouse.GetGlobalState`](https://github.com/MonoGame/MonoGame/blob/3e223ada704c8fbbe4f38613f456c515c20394fb/MonoGame.Framework/Platform/Input/Mouse.SDL.cs#L27), which is not supported on Wayland. This can probably be fixed by making MonoGame use mouse button up/down events instead.~ Fixed in https://github.com/YaLTeR/MonoGame/commit/683eedf1ec13618e263c0c85b39366a847eb635b.
- In some cases my keyboard overrides are ignored, for instance Ctrl-something shortcuts use the Ctrl button even though I have it globally bound to CapsLock. This is a SDL issue: https://bugzilla.libsdl.org/show_bug.cgi?id=4889
- ~Drag and drop onto the game doesn't work.~ Fixed in https://github.com/Quaver/MonoGame/pull/5.

Without the mouse clicks it's impossible to get into gameplay, however by hardcoding the select screen transition I'm able to get into gameplay fine and it seems to work well. So as far as my very limited testing shows, it's limited to the three issues above, plus having to bundle a newer version of libSDL2 (since the one currently bundled crashes when attempting to run in Wayland mode).

YaLTeR:
In addition to figuring out these two remaining issues (fairly certain xkb overrides need to be fixed in SDL, not sure about drag and drop yet), it would be cool to have an (on-by-default) option to prefer Wayland for Quaver so that you don't have to tinker with the SDL_VIDEODRIVER environment variable to use the Wayland VSync. The way I can see it implemented is by trying to init the Wayland video driver in SDL first, and only if that fails trying to init SDL with the default options. This requires changes to MonoGame.

PR #2057, opened 2020-05-28
Fix drag and drop on Wayland

YaLTeR:
cc #944


YaLTeR:
Or maybe I should do this in mg?

Swan:
> Or maybe I should do this in mg?

Yeah, that would save the trouble of having to do this each time

YaLTeR:
https://github.com/Quaver/MonoGame/pull/5

PR #2144, opened 2020-06-17
Reduce volume on focus loss

### LINKED ISSUE #2143: Add an option to duck game volume when game window is inactive
When tabbing out of the game, often it's because I want to listen to something (watch back a recording of my gameplay for example). It's tedious to reset my volume every time I do this.

There should be a toggle in the audio settings, which, when enabled, the game should automatically reduce the volume to a percentage of its current setting when the window loses focus, then return it to its original setting when the game window regains focus.

N/A

monotron:
Closes #2143.

This PR adds the ability to reduce the volume of the game when window focus is lost.

[Video demonstration available here.](https://streamable.com/rh4i2o)

### DIFF
--- a/Quaver.Shared/Config/ConfigManager.cs
+++ b/Quaver.Shared/Config/ConfigManager.cs
@@ -472,7 +472,7 @@ namespace Quaver.Shared.Config
         /// <summary>
         /// The percentage to scale the master volume by when the game window loses focus
         /// </summary>
-        internal static Bindable<int> VolumePercentageOnWindowInactive { get; private set; }
+        internal static BindableInt VolumePercentageOnWindowInactive { get; private set; }
 
         /// <summary>
         /// </summary>
@@ -916,7 +916,7 @@ namespace Quaver.Shared.Config
             MultiplayerLobbyVisibilityType = ReadValue(@"MultiplayerLobbyVisibilityType", MultiplayerLobbyRoomVisibility.All, data);
             UseSteamWorkshopSkin = ReadValue(@"UseSteamWorkshopSkin", false, data);
             LowerFpsOnWindowInactive = ReadValue(@"LowerFpsOnWindowInactive", true, data);
-            VolumePercentageOnWindowInactive = ReadValue(@"VolumePercentageOnWindowInactive", 90, data);
+            VolumePercentageOnWindowInactive = ReadInt(@"VolumePercentageOnWindowInactive", 100, 0, 100, data);
             DownloadDisplayOwnedMapsets = ReadValue(@"DownloadDisplayOwnedMapsets", true, data);
             DownloadReverseSort = ReadValue(@"DownloadReverseSort", false, data);
             DisplayNotificationsBottomToTop = ReadValue(@"DisplayNotificationsBottomTotop", false, data);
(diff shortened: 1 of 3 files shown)


robby250:
Cool stuff, I was thinking it could also do a transition over ~200ms or so, from the focused volume to the unfocused one so that it sounds smoother. But that's a bit more complicated I suppose

YaLTeR:
There's a bass call to do that, should be easy.

Quaver.Shared/QuaverGame.cs
```diff
@@ -185,6 +185,31 @@ public string Version
...
+            get => windowUnfocused;
+            set
+            {
+                if (value != windowUnfocused)
+                {
+                    windowUnfocused = value;
+                    WindowFocusChanged?.Invoke(this, null);
+                }
+            }
+        }
```
Swan, line 206: This event should be implemented at [Framework level](https://github.com/Quaver/Wobble/blob/master/Wobble/WobbleGame.cs), so that all games that use Wobble can take advantage of it.

Quaver.Shared/QuaverGame.cs
```diff
@@ -185,6 +185,31 @@ public string Version
         /// </summary>
         private bool WindowActiveInPreviousFrame { get; set; }
 
+        private event EventHandler WindowFocusChanged;
```
Swan, line 188: Instead of this, I would create a special type of EventArgs for this which contains the new focus value. It would allow subscribers of the event to have access to the new value from anywhere.

```cs
private event EventHandler<FocusChangeEventArgs> WindowFocusChanged;
```



```cs
public class FocusChangeEventArgs : EventArgs
{
    public bool IsFocused { get; }

    public FocusChangeEventArgs(bool isFocused) => IsFocused = isFocused;
}
```

Quaver.Shared/QuaverGame.cs
```diff
@@ -819,6 +847,36 @@ private void LimitFpsOnInactiveWindow()
...
+        /// </summary>
+        /// <param name="sender">The object which triggered the event.</param>
+        /// <param name="e">Extra event arguments.</param>
+        private void ScaleVolumeOnFocusChange(object sender, EventArgs e)
+        {
+            if (WindowUnfocused)
+            {
+                var newVolume = (int) (ConfigManager.VolumeGlobal.Value * (ConfigManager.VolumePercentageOnWindowInactive.Value / 100.0f));
+                OriginalVolume = ConfigManager.VolumeGlobal.Value;
+                ConfigManager.VolumeGlobal.Value = newVolume;
```
Swan, line 872: Changing a user's config to update the volume is not a good solution.

Try changing `AudioTrack.GlobalVolume` and `AudioSample.GlobalVolume` instead. This way you won't overwrite the user's preferred setting (in the event of the game closing with the config being saved back to the original value).

Quaver.Shared/Config/ConfigManager.cs
```diff
@@ -911,6 +916,7 @@ private static void ReadConfigFile()
             MultiplayerLobbyVisibilityType = ReadValue(@"MultiplayerLobbyVisibilityType", MultiplayerLobbyRoomVisibility.All, data);
             UseSteamWorkshopSkin = ReadValue(@"UseSteamWorkshopSkin", false, data);
             LowerFpsOnWindowInactive = ReadValue(@"LowerFpsOnWindowInactive", true, data);
+            VolumePercentageOnWindowInactive = ReadInt(@"VolumePercentageOnWindowInactive", 100, 0, 100, data);
```
Swan, line 919: The default value should probably be something low like 20 or 30.

monotron:
then I'll close this PR for now, raise a new one in Wobble and approach it again :)

Issue #1400, opened 2020-01-02
[Skinning] Column specific ColumnSize variables


robby250:
Add the option to customize each column with a specific ColumnSize in a way that doesn't affect the current skins
Examples (for 4k):
- ColumnSize:40 would make all 4 columns 40 size.
- ColumnSize:40,35 would alternate the column size: 40,35,40,35
- ColumnSize:40,35,36,37 would change each column respectively

So in essence it should just loop the columnsize values in order until each column has a columnsize.

LevFlavien:
Tbh this is the only missing feature that prevents me from playing Quaver :smile:

PR #2234, opened 2020-07-04
Column size per lane

Adrriii:
https://github.com/Quaver/Quaver/issues/1400

Still needs work


robby250:
In my example, the 2 values alternating was just an example for how the algorithm should work, not a specific case.
It should use the same pattern to fill the ColumnSize array no matter how many values are provided.
[7K]
ColumnSize = 50,53,58 => The columns are filled up as 50,53,58,50,53,58,50 (and 53 for the scratch)

If there are more values provided than what it needs, then it just picks the first ones.
[7K]
ColumnSize = 51,52,53,54,55,56,57,58,59,60 => The columns are: 51,52,53,54,55,56,57 (and 58 for the scratch). It should ignore the values that aren't needed.

Also just a reminder, for the preview and wherever else the ColumnSize is scaled, it should divide the size of the preview with the values in the ColumnSize array added up to get the ratio between the sizes of the two, and then scale each value in the array down with the ratio, in order to preserve the column ratios.

Example:
[4K]
ColumnSize = 50, 55, 55, 50 => added up it's 210
previewSize = 160

ratio = 160/210 = 0.7619...
new ColumnSize for preview = 50 * ratio, 55 * ratio, 55 * ratio, 50 * ratio

Adrriii:
I don't think allowing incorrectly formatted values is in the spirit of quaver's skinning, giving multiple possibilities for the same value is already weird looking in the code and is makes room for errors

robby250:
It's just versatile code lol, it's supposed to do just one thing which is loop the ColumnSize value(s) in order until all the columns have a value and then stop. I don't understand what you mean by multiple possibilities for the same value

Adrriii:
The problem is not the code, it's the skin configuration. If you put 3 values, I wouldn't expect it to use the first for the fourth column.
This is, imo, a very complex structure just to allow a very lazy (and ugly) configuration.
Using one value for everything or specify each columns is the only pair that makes sense, I added the alternating stuff because I could but it's honestly not looking good

robby250:
Ok then make it only possible to have 1 value or 4/7(8 for 7k+1 lol) values so that each column is assigned a value. The thing I proposed was something generic that would work with any amount of values, be it 1 or 8 or anything in between.

Adrriii:
I'm not familiar with the scratch lane configuration, is it an entire new category or is it inside 7k ?
If it's inside 7k, should it be on the left or on the right ?

robby250:
In skinning it's basically the 8th key, but I'm not sure in code.

PR #2249, opened 2020-07-08
Option for ranked accuracy to be displayed even with custom judges

Adrriii:
I find this feature very interesting from etterna, useful when playing harder judges on a daily basis


Adrriii:
reminder: check if it works in multiplayer

PR #2255, opened 2020-07-09
Only disable HardwareModeSwitch on Linux

Swan:


YaLTeR:
We use a different os check in other places?

Swan:
Right. Apparently `RuntimeInformation.IsOSPlatform()` is more reliable.

img

Issue #2021, opened 2020-05-21
Don't crash on startup if Steam auth ticket is stale


YaLTeR:
**Describe the bug**
When Steam auth ticket is stale (for example, this can happen on unstable internet connections or when switching internet connection e.g. from one wi-fi to another) Quaver crashes on startup with:
```
[4:45:57] - NETWORK - ERROR: Could not log in because the steam auth session ticket was not validated.
[4:45:57] - RUNTIME - ERROR: System.Exception: Failed to login
   at Quaver.Shared.Online.OnlineManager.Login() in /home/yalter/source/cs/Quaver/Quaver.Shared/Online/OnlineManager.cs:line 209
   at Quaver.Shared.Screens.Main.MainMenuScreen.OnFirstUpdate() in /home/yalter/source/cs/Quaver/Quaver.Shared/Screens/Main/MainMenuScreen.cs:line 76
   at Quaver.Shared.Screens.QuaverScreen.Update(GameTime gameTime) in /home/yalter/source/cs/Quaver/Quaver.Shared/Screens/QuaverScreen.cs:line 80
   at Quaver.Shared.Screens.Main.MainMenuScreen.Update(GameTime gameTime) in /home/yalter/source/cs/Quaver/Quaver.Shared/Screens/Main/MainMenuScreen.cs:line 87
   at Wobble.Screens.ScreenManager.Update(GameTime gameTime) in /home/yalter/source/cs/Quaver/Wobble/Wobble/Screens/ScreenManager.cs:line 86
   at Wobble.WobbleGame.Update(GameTime gameTime) in /home/yalter/source/cs/Quaver/Wobble/Wobble/WobbleGame.cs:line 196
   at Quaver.Shared.QuaverGame.Update(GameTime gameTime) in /home/yalter/source/cs/Quaver/Quaver.Shared/QuaverGame.cs:line 347
   at Microsoft.Xna.Framework.Game.DoUpdate(GameTime gameTime) in /home/yalter/source/cs/Quaver/Wobble/MonoGame/MonoGame.Framework/Game.cs:line 678
   at Microsoft.Xna.Framework.Game.Tick() in /home/yalter/source/cs/Quaver/Wobble/MonoGame/MonoGame.Framework/Game.cs:line 525
[...]

PR #2276, opened 2020-07-11
Quick fix for auto login crash when offline

lolepop:
Game currently crashes when launched with steam offline or as described in #2021 when launched with the auto login setting. No auto retrying since idk how game dev works lol.


lolepop:
Fixed. Notepad++ might not have been the best ide after all

PR #2410, opened 2020-07-16
Waveform in the editor

VortexCoyote:
implemented a waveform for the editor, which is also toggleable through the "view" tab, while it's state is saved to the config file.

### DIFF
--- a/Quaver.Shared/Screens/Edit/EditScreen.cs
+++ b/Quaver.Shared/Screens/Edit/EditScreen.cs
@@ -150,6 +150,10 @@ namespace Quaver.Shared.Screens.Edit
 
         /// <summary>
         /// </summary>
+        public Bindable<bool> ShowWaveform { get; } = ConfigManager.EditorShowWaveform ?? new Bindable<bool>(true) {Value = true};
+
+        /// <summary>
+        /// </summary>
         public Bindable<EditorBeatSnapColor> BeatSnapColor { get; } = ConfigManager.EditorBeatSnapColorType ?? new Bindable<EditorBeatSnapColor>(EditorBeatSnapColor.Default);
 
         /// <summary>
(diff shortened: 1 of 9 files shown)


Quaver.Shared/Screens/Edit/EditScreen.cs
```diff
@@ -374,13 +381,16 @@ private void SetAudioTrack(IAudioTrack track)
                 Track = track;
                 Track.ApplyRate(ConfigManager.Pitched?.Value ?? true);
 
-                lock (AudioEngine.Track)
+                if (AudioEngine.Track != null)
```
YaLTeR, line 384: Why have you moved the lock inside?

Quaver.Shared/Screens/Edit/UI/Playfield/Waveform/EditorPlayfieldWaveform.cs
```diff
@@ -0,0 +1,114 @@
...
+                        TrackSliceData[y, 0] = TrackData[index];
+                        TrackSliceData[y, 1] = TrackData[index + 1];
+                    }
+                }
+
+                var slice = new EditorPlayfieldWaveformSlice(Playfield, SliceSize, TrackSliceData, t);
+                Slices.Add(slice);
+            }
+
+            Bass.StreamFree(Stream);
```
YaLTeR, line 61: This looks like it belongs in Destroy

Quaver.Shared/Screens/Edit/EditScreen.cs
VortexCoyote: Yes, because it crashed in Visual Tests and Swan told me to make that change

Quaver.Shared/Screens/Edit/UI/Playfield/Waveform/EditorPlayfieldWaveform.cs
VortexCoyote: Sure thying

Issue #2319, opened 2020-07-12
Workflow improvements for SVs in Editor


Kanarisky:
**Is your feature request related to a problem? Please describe.**
While Quaver's editor is fantastic, there are some possible workflow improvements to be made for SVs.
<br />

**1. SVs should be visible in the density graph, or next to it.** 
It would be easier to manage your SVs scattered throughout the map. Something like this, either on the side of—or within—the density graph:
![image](https://user-images.githubusercontent.com/68185248/87245142-60fbd500-c486-11ea-91fe-c56989b654e6.png)
<br />

**2. A hotkey for creating a new SV point.** 
Preferably rebindable. This is *ctrl + shift + P* in Osu. With this you can place many points sequentially, then *ctrl + click* every 2nd or 3rd one in the SV window to apply pattern SV in bulk. Currently, adding SVs is very slow.
<br />

**3. Mouse wheel / arrow key control of the notefield while mouse is within the non-list section of the SV window.** 
Again, to make it quicker to add SVs. Currently you must have the mouse off the SV window, scroll to the next tick, mouse over *Add* and click it, then mouse off the SV window and scroll another tick, etc. 

The above also applies to left and right arrow keys; it's based on your mouse position. In fact, if you're in the middle of typing in the "Multiplier" entry field with your mouse off the SV window and you press left or right arrow keys, it will both shift the typing position, and move forward/backward a tick in the notefield. I believe this is a bug.

With your mouse within the green rectangle below you should still have scroll wheel functionality, but you currently don't:
![image](https://user-images.githubusercontent.com/68185248/87245304-b5ec1b00-c487-11ea-86c5-ff2f28aeee9e.png)
<br />

**4. Automatically select the string in "Multiplier" entry field upon creation of an SV point.** 
Makes it quicker to create SVs, rather than having to click onto the field for every single new SV.
![image](https://user-images.githubusercontent.com/68185248/87245407-50e4f500-c488-11ea-9bdf-8616ee084f32.png)
<br />

**5. Allow navigation of SV points in the list with up/down arrow keys while the SV window is in focus.**
(Not necessarily while your mouse is hovering the SV window)
<br />

**6. While an SV point is selected in the list, pressing *Enter* should select the string in "Multiplier" entry field.** 
Makes editing points quicker, and doable completely with keyboard rather than swapping from keyboard to mouse repeatedly.
<br />

[...]

PR #2419, opened 2020-07-17
SV workflow improvements

IceDynamix:
Partially addresses #2319 

- Make the panel vertically resizable
- Range selection with Shift+Click
- Add "Select current SV" button to jump to the current SV (as in current editor timestamp) in the list and add it to the current selection
- Select the "Multiplier" field when creating a new SV
- Deselect all keybind (Ctrl+D)
- All changes also applied to timing points panel

### DIFF
--- a/Quaver.Shared/Screens/Edit/Plugins/Timing/EditorScrollVelocityPanel.cs
+++ b/Quaver.Shared/Screens/Edit/Plugins/Timing/EditorScrollVelocityPanel.cs
@@ -209,8 +209,9 @@ namespace Quaver.Shared.Screens.Edit.Plugins.Timing
                 var currentPoint = Screen.WorkingMap.GetScrollVelocityAt(Screen.Track.Time);
                 if (currentPoint != null)
                 {
-                    SelectedScrollVelocities.Add(currentPoint);
                     NeedsToScrollToLastSelectedSv = true;
+                    if (!SelectedScrollVelocities.Contains(currentPoint))
+                        SelectedScrollVelocities.Add(currentPoint);
                 }
             }
 
--- a/Quaver.Shared/Screens/Edit/Plugins/Timing/EditorScrollVelocityPanel.cs
+++ b/Quaver.Shared/Screens/Edit/Plugins/Timing/EditorScrollVelocityPanel.cs
@@ -92,9 +92,9 @@ namespace Quaver.Shared.Screens.Edit.Plugins.Timing
         /// </summary>
         protected override void RenderImguiLayout()
         {
-            ImGui.SetNextWindowSize(new Vector2(356, 480));
+            ImGui.SetNextWindowSizeConstraints(new Vector2(356, 0), new Vector2(356, float.MaxValue));
             ImGui.PushFont(Options.Fonts.First().Context);
-            ImGui.Begin(Name, ImGuiWindowFlags.NoResize);
+            ImGui.Begin(Name);
 
             DrawHeaderText();
             ImGui.Dummy(new Vector2(0, 10));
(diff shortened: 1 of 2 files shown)


AntecMoma:
Now this is one big PP moment.

AntecMoma:
Also make the SV easier to use as copy and past SV values. and selecting via: Select from: MOUSE1 Select to: CTRL+SHIFT+MOUSE1  this can also be used to select rows not just SV and BPM values,

IceDynamix:
It was already possible to use SVs/BPMs as copy paste values via Ctrl+X/C/V (the Ctrl is optional actually). You just have to hover over the panel while executing the keybinds (including Ctrl+D). It will paste at the current editor position by default.

Quaver.Shared/Screens/Edit/Plugins/Timing/EditorTimingPointPanel.cs
```diff
@@ -186,7 +192,31 @@ private void DrawRemoveButton()
...
+        /// <summary>
+        /// </summary>
+        private void DrawGoToCurrentTimingPointButton()
+        {
+            if (ImGui.Button("Go to current timing point"))
+            {
+                var currentPoint = Screen.WorkingMap.GetTimingPointAt(Screen.Track.Time);
+                if (currentPoint != null)
+                {
+                    SelectedTimingPoints.Add(currentPoint);
```
Swan, line 208: This should probably clear the currently selected timing points before adding the new one? Otherwise it'll select multiple. Some goes for both panels.

Quaver.Shared/Screens/Edit/Plugins/Timing/EditorScrollVelocityPanel.cs
```diff
@@ -92,13 +96,15 @@ public void Initialize()
...
+            ImGui.SetNextWindowSizeConstraints(new Vector2(356, 0), new Vector2(356, float.MaxValue));
             ImGui.PushFont(Options.Fonts.First().Context);
-            ImGui.Begin(Name, ImGuiWindowFlags.NoResize);
+            ImGui.Begin(Name);
 
             DrawHeaderText();
             ImGui.Dummy(new Vector2(0, 10));
 
+            DrawGoToCurrentSVButton();
+
```
Swan, line 107: Would be good to add 10px spacing here. Same for both panels.

Quaver.Shared/Screens/Edit/Plugins/Timing/EditorTimingPointPanel.cs
IceDynamix: That's intended and documented in the tooltip a few lines below. The reason for that is to make selecting a range easier. One scenario would be going to a point and selecting it, scrolling to the end of the section you want to select in the editor and clicking the "Select SV". You can then unselect and Shift+Click it again to select the entire range. Another scenario is simply to scroll to different locations in the map and selecting each SV by scrolling to it in the editor and clicking the button, removing the need to figure out which one is which just from the numbers in the table alone. I should probably clarify that intention in the tooltip. Another way to clarify the intention would be to add a "n SV points selected" label somewhere.

Swan: The button itself could just be phrased a bit better if it has this behavior. For instance, "Go To Objects" in the editor means that every other object gets deselected and the new ones are selected.

Quaver.Shared/Screens/Edit/Plugins/Timing/EditorScrollVelocityPanel.cs
```diff
@@ -186,7 +196,31 @@ private void DrawRemoveButton()
...
+        /// <summary>
+        /// </summary>
+        private void DrawSelectCurrentSVButton()
+        {
+            if (ImGui.Button("Select current SV"))
+            {
+                var currentPoint = Screen.WorkingMap.GetScrollVelocityAt(Screen.Track.Time);
+                if (currentPoint != null)
+                {
+                    SelectedScrollVelocities.Add(currentPoint);
```
Swan, line 212: Add a check if it's already in the list before adding it. This will add multiple of the same object if you keep clicking it. Same applies for both panels.

PR #2502, opened 2020-07-29
Rework Game Initialization

### LINKED ISSUE #891: Add a splash screen to make it clear quaver is launching
When launching the quaver client, it tends to start as a background process, and even not launch at all.

 Adding a splash screen that could be as simple as a loading wheel with the quaver logo would make it easy to determine if quaver crashed on launch, if it never launched, or if it's currently launching. 

The only alternative is better reporting to the steam client that the game is launching (which is a limitation of the client), or telling users to check task manager for the QuaverGame.exe process.

Swan:
Closes #891

### DIFF
--- a/Quaver/Program.cs
+++ b/Quaver/Program.cs
@@ -44,29 +44,7 @@ namespace Quaver
         private static int IpcPort { get; } = 43596;
 
         [STAThread]
-        public static void Main(string[] args)
-        {
-            // Prevents more than one instance of Quaver to run at a time
-            using(var mutex = new Mutex(false, "Global\\" + Guid))
-            {
-                if(!mutex.WaitOne(0, false))
-                {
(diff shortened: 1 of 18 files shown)


Quaver/Program.cs
```diff
@@ -44,29 +44,7 @@ public static class Program
...
-            // Prevents more than one instance of Quaver to run at a time
-            using(var mutex = new Mutex(false, "Global\\" + Guid))
-            {
-                if(!mutex.WaitOne(0, false))
-                {
-                    Console.WriteLine("Quaver is already running");
-
-                    // Send to running instance only if we have actual data to send
-                    if (args.Length > 0)
-                        SendToRunningInstanceIpc(args);
```
YaLTeR, line 58: So how is this done after this commit?

PR #2560, opened 2020-08-15
Add more editor keybinds

IceDynamix:
Implemented:
- Number row to place/remove notes
- Ctrl+I to place a SV, Ctrl+Shift+I to place a timing point
- Shift+Up/Down to switch between layers
- Home/End to seek to the start/end of map
- Shift+H to hide the current layer

Open for better keybind alternatives


cha0sbuster:
Yes! I was just thinking about how helpful being able to input with the number row like with ArrowVortex would be.

PR #2567, opened 2020-08-17
Ui redesign online

NekodNekod:
Can we get a global key to pause and unpause the music playing? thank you


Swan:
CTRL+P works everywhere

PR #2570, opened 2020-08-17
Add "Resnap" editor functions

IceDynamix:
Adds two options in the editor under Edit > Resnap all notes, one to resnap to the current snap and one to resnap to 1/16 AND 1/12 by default (without using a common multiple). This will also resnap LN ends. Undo structure is also available.

### DIFF
--- a/Quaver.Shared/Screens/Edit/Actions/HitObjects/Resnap/EditorActionResnapHitObjects.cs
+++ b/Quaver.Shared/Screens/Edit/Actions/HitObjects/Resnap/EditorActionResnapHitObjects.cs
@@ -29,23 +29,26 @@ namespace Quaver.Shared.Screens.Edit.Actions.HitObjects.RemoveBatch
         private List<int> Snaps { get; }
 
         /// <summary>
-        ///     Unmodified notes, kept for undoing
         /// </summary>
-        private List<HitObjectInfo> OldNotes { get; } = new List<HitObjectInfo>();
+        private List<HitObjectInfo> HitObjectsToResnap { get; }
 
         /// <summary>
-        ///     Modified notes, kept for undoing
+        ///  The distance of note start/end to the closest snap
         /// </summary>
-        private List<HitObjectInfo> NewNotes { get; } = new List<HitObjectInfo>();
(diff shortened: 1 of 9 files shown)


Quaver.Shared/Screens/Edit/Actions/HitObjects/Resnap/EditorActionResnapHitObjects.cs
```diff
@@ -0,0 +1,128 @@
...
+
+        /// <inheritdoc />
+        /// <summary>
+        /// </summary>
+        public void Undo()
+        {
+            NewNotes.ForEach(n => WorkingMap.HitObjects.Remove(n));
+            WorkingMap.HitObjects.AddRange(OldNotes);
+            WorkingMap.Sort();
+        }
```
Swan, line 126: Few issues with this:

1. Undoing and redoing adds the same notes to the list and more objects to the map - img
2. The event for this action isn't handled in `EditorPlayfeld`, so notes that are moved by this do not update.
3. The event for this action isn't handled in `EditorMapPreview`, so performing this action will not refresh the gameplay preview.
4. The event for this action isn't handled in `EditorPanelDetails`, so the object count & difficulty rating are never updated.
5. The event isn't nullified on `Dispose()` in `EditorActionManager` - causing a memory leak.

IceDynamix:
Update: Stuck on resolving the EditorPlayfield updating, since, even after assigning all event handlers and confirming they are executed (with a debug console), the note positions aren't updated.

Swan:
> Update: Stuck on resolving the EditorPlayfield updating, since, even after assigning all event handlers and confirming they are executed (with a debug console), the note positions aren't updated.

This is because you're creating entirely new objects with a deep clone, so the playfield doesn't have references to the new objects that you're creating.

Rather than creating new objects and getting rid of the old ones, just change the times of the already existing objects, and it should work.

[See here](https://github.com/Quaver/Quaver/pull/2570/commits/5982a4000675e8f36aeb524f7784c747978741f8#diff-ad1aba0c274858067c6b37634554999aR84-R91)

Swan (approved):

PR #2618, opened 2020-09-02
Fix cursor disappearing when trying to testplay with invalid map

### LINKED ISSUE #1908: [Editor] Cursor disappears when trying (and failing) to testplay
Cursor disappears in the editor, if you try to testplay a map from a point, after which there are no objects placed. The dialog "There are no hitobjects to play past this point" shows up as intended, but at the same time the cursor disappears. I assume this happens because the game is setting you up for gameplay screen, where cursor is intended to be invisible.

Illuminati-CRAZ:
Fixes #1908


Illuminati-CRAZ:
Whoops, accidentally had commits of other branch in this.

PR #2733, opened 2020-11-12
Changes related to editor layers

### LINKED ISSUE #620: Editor: Make merging layers more UX friendly
Preferable keybind would be Ctrl+E like in photoshop.
It should be possible to merge the current layer to the layer below without having to select a layer.
It should also be possible to merge 3 or more layers together.

### LINKED ISSUE #2595: Adding notes with 1-4 keys is always added to the Default Layer
**Describe the bug**
As the title says. It also happened in 7k.

**To Reproduce**
1. Open Editor in Quaver
2. Select/ add other layer.
3. Adding notes with 1-4 keys

**Expected behavior**
It should be added to the layer that the mapper used to add.

Illuminati-CRAZ:
Additions/Changes
- Make toggling layer visibility not add to the undo stack
- Implement inserting an editor layer at a given index
- Creating a new layer using the editor now creates a new layer under the currently selected layer
- Plugins can specify an index to insert a layer at
- Top row number keys now add notes to selected layer (Fixes #2595)
- Implement layer merging (Fixes #620)
  - Ctrl+M to merge current layer into the one above
  - Ctrl+Alt+M to merge current layer into the default layer

Code Stuff
- Handling the index might be jank
- Updating the DrawableEditorLayers might be jank
- Made EditorLayerRemovedEventArgs and EditorLayerRenamedEventArgs not inherit from EditorLayerCreatedEventArgs
- ActionManager.EditScreen changed to public

Problems
- EditorActionMergeLayers unintuitively selects the layer above the one removed
- Merging through the right click menu is unintuitive
- Will need to add the layer merge action to the plugin utility function for creating actions eventually

To do
- [ ] Add move objects to layer to right click menu
- [ ] Change it so SelectedLayer.Value is set to DefaultLayer instead of null when Default Layer is selected

### DIFF
--- a/Quaver.Shared/Graphics/Containers/PoolableScrollContainer.cs
+++ b/Quaver.Shared/Graphics/Containers/PoolableScrollContainer.cs
@@ -230,14 +230,39 @@ namespace Quaver.Shared.Graphics.Containers
             {
                 var drawable = CreateObject(AvailableItems[index], index);
                 drawable.DestroyIfParentIsNull = false;
-                drawable.Y = (PoolStartingIndex + index) * drawable.Height + PaddingTop;
+
+                Pool.Insert(index, drawable);
 
                 if (updateContent)
-                    drawable.UpdateContent(AvailableItems[index], index);
-
-                Pool.Add(drawable);
+                {
+                    for (int i = index; i < Pool.Count; i++)
+                    {
+                        var foo = Pool[i];
+                        foo.Y = (PoolStartingIndex + i) * foo.Height + PaddingTop;
+                        foo.UpdateContent(AvailableItems[i], i);
+                    }
+                }
 
                 return drawable;
+            }
+        }
+
+        protected void AddObjectAtIndex(int index, T obj, bool scrollTo, bool usePoolCount = false)
+        {
+            lock (AvailableItems)
+            lock (Pool)
+            {
+                if (!AvailableItems.Contains(obj))
+                    AvailableItems.Insert(index, obj);
+
+                // Need another drawable to use
+                if (Pool.Count < PoolSize)
+                    AddContainedDrawable(AddObject(index));
+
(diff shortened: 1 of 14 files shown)


IceDynamix:
Why should layer visibility not add to the undo stack?

Illuminati-CRAZ:
Personally I feel it doesn't really change the map in a meaningful way.
Layer visibility feels more like an visual preference than part of the map, and undoing preferences is weird.

(Also I'm making a plugin and it changes layer visibility often and makes undoing impossible if toggling layer visibility contributes to the undo stack)

Illuminati-CRAZ:
also iirc, changing layer visibility originally didn't add to the undo stack, and does so now because it's now an editor action

Illuminati-CRAZ:
Not sure whether a shortcut for merging layers should be added.
It'd probably be Ctrl+M, not sure whether to merge up or merge into default layer.

Illuminati-CRAZ:
@Swan Is there a reason why SelectedLayer.Value is set to null and not DefaultLayer when the Default Layer is selected?

Swan:
> @Swan Is there a reason why SelectedLayer.Value is set to null and not DefaultLayer when the Default Layer is selected?

I don't remember fully, but it's most likely because the default layer isn't addressed anywhere in the .qua file, so it was written to just be null. You technically could set it to the default layer, and it should still work fine (test and see if the file is still written the same way).

Swan:
It's worth mentioning that all of this should be in separate PRs. Extremely large PRs that are adding/changing multiple things at a time are something we do not accept.

From the [README](https://github.com/Quaver/Quaver#contributing):

> Any contributions can be made through opening a pull request. When contributing however, please keep in mind that there should only be one branch/pull request per feature. This means if your branch is titled my-new-feature, you shouldn't have any unrelated commits, and the same system is applied to pull requests. Please make sure to keep your pull requests short and concise.

PR #2740, opened 2020-11-15
Implement EditorActionBatch to group actions into a single action

Illuminati-CRAZ:
Mainly for plugins

- [x] comment code

- [x] fix weird table -> list -> object -> list thing not working in util function
       turns out it was actually dynvalue -> table -> object -> table -/-> list or something along those lines


Illuminati-CRAZ:
The util function for creating an EditorAction breaks when a plugin sends a table/list to it and idk why

Illuminati-CRAZ:
Figured it out.
The conversion from MoonSharp.Interpreter.DynValue storing lua table data into a C# object results in it being converted into a MoonSharp.Interpreter.Table, which I can't figure out a way to easily convert into a list.

Solution is to have utility function keep arguments as DynValues and within the function convert them into the appropriate type using DynValue.ToObject()

I hope this isn't word spaghetti

Quaver.Shared/Screens/Edit/Plugins/EditorPluginUtils.cs
```diff
@@ -125,71 +125,68 @@ public static EditorLayerInfo CreateEditorLayer(string name, bool hidden = false
...
         /// </summary>
         /// <param name="type"></param>
-        /// <param name="arg1"></param>
-        /// <param name="arg2"></param>
-        /// <param name="arg3"></param>
-        /// <param name="arg4"></param>
+        /// <param name="args"></param>
         /// <returns></returns>
-        public static IEditorAction CreateEditorAction(EditorActionType type, DynValue arg1 = null, DynValue arg2 = null, DynValue arg3 = null, DynValue arg4 = null)
+        public static IEditorAction CreateEditorAction(EditorActionType type, params DynValue[] args)
```
Swan, line 130: Is there any way to avoid having this method? (Like creating the objects directly in Lua after they're exposed)

Illuminati-CRAZ: I tried finding another way to do this, but as far as I can tell this is pretty much the only way to do it

Swan (approved):

PR #2744, opened 2020-11-18
Fix LN rendering past the receptor if hit late

Kizuna-dev:
Remove the "**[CurrentLongNoteBodySize + LongNoteSizeDifference <= HitObjectSprite.Height / 2f]**" line, which block the LN Body/Head from being rendered if the height was below half the hitobject's height. As this is an unnatural behavior i don't think this is necessary to keep.

Add "**[curTime >= Info.EndTime && CurrentlyBeingHeld]**", whose purpose is to fix the LN going beyond the hitposition if hit after the endtime AND only happen when trying to hit the object (otherwise every LN would dissapear once the current time is the endtime)

For example, currently LN act as such : https://streamable.com/0orq25 in which you can notice the LN body goes beyond the receptor cause hit too late.

And with the fix : https://streamable.com/qu1tln the LN doesn't go past the receptor if hit late

Also i'm not sure by what to change the comment so feel free to fix that one

### DIFF
--- a/Quaver.Shared/Screens/Gameplay/Rulesets/Keys/HitObjects/GameplayHitObjectKeys.cs
+++ b/Quaver.Shared/Screens/Gameplay/Rulesets/Keys/HitObjects/GameplayHitObjectKeys.cs
@@ -449,9 +449,9 @@ namespace Quaver.Shared.Screens.Gameplay.Rulesets.Keys.HitObjects
 
             LongNoteEndSprite.Y = GetEndSpritePosition(offset, EndTrackPosition);
 
-            // Stop drawing LN body + end if the ln reaches half the height of the hitobject
+            // Stop drawing LN body + end if the ln reaches 0 height or below
             // (prevents body + end extending below this point)
-            if (CurrentLongNoteBodySize + LongNoteSizeDifference <= HitObjectSprite.Height / 2f || CurrentLongNoteBodySize <= 0)
+            if (CurrentLongNoteBodySize <= 0 || curTime >= Info.EndTime && CurrentlyBeingHeld)
             {
                 LongNoteEndSprite.Visible = false;
                 LongNoteBodySprite.Visible = false;
--- a/Quaver.Shared/Screens/Gameplay/Rulesets/Keys/HitObjects/GameplayHitObjectKeys.cs
+++ b/Quaver.Shared/Screens/Gameplay/Rulesets/Keys/HitObjects/GameplayHitObjectKeys.cs
@@ -449,9 +449,9 @@ namespace Quaver.Shared.Screens.Gameplay.Rulesets.Keys.HitObjects
 
             LongNoteEndSprite.Y = GetEndSpritePosition(offset, EndTrackPosition);
 
-            // Stop drawing LN body + end if the ln reaches 0 height or below
+            // Stop drawing LN body + end if the ln reaches half the height of the hitobject
             // (prevents body + end extending below this point)
-            if (CurrentLongNoteBodySize <= 0 || curTime >= Info.EndTime && CurrentlyBeingHeld)
+            if (CurrentLongNoteBodySize + LongNoteSizeDifference <= HitObjectSprite.Height / 2f || CurrentLongNoteBodySize <= 0 || curTime >= Info.EndTime && CurrentlyBeingHeld)
             {
                 LongNoteEndSprite.Visible = false;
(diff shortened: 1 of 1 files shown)


YaLTeR:
> Remove the "CurrentLongNoteBodySize + LongNoteSizeDifference <= HitObjectSprite.Height / 2f" line, which block the LN Body/Head from being rendered if the height was below half the hitobject's height. As this is an unnatural behavior i don't think this is necessary to keep.

The commit that added this is https://github.com/Quaver/Quaver/commit/865a74478255e9d3821474fb8d441821454ae30e#diff-bf5bd1ef67286ac65eb752a1fabd2699bea48dd242352c815cc0a51d4b0296af. Could you please demonstrate that you haven't broken what that commit was trying to fix? It's not obvious to me.

> curTime >= Info.EndTime && CurrentlyBeingHeld

Please enclose it in brackets, || and && precedence is not obvious.

> video examples

I'm not sure what I'm looking for here. Showing it with a very low scroll speed may help.

Have you tested this change with all types of skins (bars, circles, arrows) as well as upscroll and downscroll? For instance, wouldn't this change make bar ends disappear abruptly?

Kizuna-dev:
> The commit that added this is [865a744#diff-bf5bd1ef67286ac65eb752a1fabd2699bea48dd242352c815cc0a51d4b0296af](https://github.com/Quaver/Quaver/commit/865a74478255e9d3821474fb8d441821454ae30e#diff-bf5bd1ef67286ac65eb752a1fabd2699bea48dd242352c815cc0a51d4b0296af). Could you please demonstrate that you haven't broken what that commit was trying to fix? It's not obvious to me.

Prior to that commit, the LNbody and LNend were going beyond the receptor/hitposition if hold beyond the endtime, which is what it was trying to fix. But it create one problem, because of this solution, which is to stop rendering the LN if it's height is below half the hitobject, all the LN below that simply doesn't render, which create situation in which you can see a LNhead without any LNbody while playing. Second problem is, that solution doesn't fix the LN being rendered past the hitposition if it's hit past the entime, recreating the previous problem.

So the solution i propose fix both of them, by removing the line i allow small LN to be rendered, and by adding **[curTime >= Info.EndTime && CurrentlyBeingHeld]**, i make sure the LN doesn't go beyond the hitposition once it is at it's entime.

In more detail, it check if current time is over or equal to the endtime and if the LN is being held, which make it so you press the LN past it's endtime, it will not render below because the current time is bigger than the endtime. As for the LN going below if the LNhead is hit too late and after the LNend, since the current time is bigger than the endtime, it also doesn't render back. 

Also, the confirmation about the LN needing to be held is also to avoid other LN just dissapearing out of the blue if it was not hit in the first place and the endtime got past the current time.

> I'm not sure what I'm looking for here. Showing it with a very low scroll speed may help.

Look closely those video, the notes are LN, as you can see i'm hitting them late, but because of them being small, i can hit the LNhead even after the LNend going beyond the hitposition, which create this effect of LN popping out below.

Here is 3 more video, 2 arrows skin (1 upscroll, 1 downscroll), and 1 bar skin. is it actually impossible for me to show such behavior with the default skin because of their hitposition being literally at the bottom of the screen, so this problem is actually not noticeable with them. And obviously bar skin have image over them so can't even notice.

https://streamable.com/fz22px
https://streamable.com/gzdi6m
https://streamable.com/ev15np

YaLTeR:
> Here is 3 more video

This is much clearer. Is that before the patch?

Kizuna-dev:
> This is much clearer. Is that before the patch?

Yes, my PR fix this behavior

Swan:
@Kizuna-dev What does your skin look like? This PR effectively reverts the functionality of the other commit.

**Before**

qw7F3Uyl8f

**After (Your PR)**

0nu6xa8tre

Kizuna-dev:
image

Just this. Tho i wanna mention that the removing of the line **"[CurrentLongNoteBodySize + LongNoteSizeDifference <= HitObjectSprite.Height / 2f]"** was because it felt unatural at some point because it block really small LN to be rendered, but if you feel like keeping it it's fine. My biggest concern is what the line **"[curTime >= Info.EndTime && CurrentlyBeingHeld]"** try to fix.

Kizuna-dev:
Re-added **"[CurrentLongNoteBodySize + LongNoteSizeDifference <= HitObjectSprite.Height / 2f]"** and added back the previous text

PR #2753, opened 2020-11-25
Fix osu songs not loading if beatmap directory is custom

### LINKED ISSUE #258: The option to load maps from osu! database file breaks the game when the Beatmap Directory setting in osu! is not default
**Describe the bug**
Quaver was unable to read the beatmap files when I tried to use the option `Load Maps From Other Games` with my osu! install. The reason is that I have the database file (osu!.db) in the other directory than osu!'s Songs directory.

**To Reproduce**
1. Go to osu!'s config and set `BeatmapDirectory` value to point somewhere else than osu!'s default directory
2. Make sure osu! works properly with that setting changed. Fill it with beatmaps and such.
3. Set a path to the osu!.db file in Quaver's config and enable `Load Maps From Other Games` option in Quaver.
4. Restart the game and watch it break being unable to read the beatmap files.

**Expected behavior**
Quaver should detect the where is the osu!'s actual beatmap directory located and read the maps from there.

**Screenshots**
https://cdn.discordapp.com/attachments/525409883877539860/525458562994929671/unknown.png

https://cdn.discordapp.com/attachments/525409883877539860/525458800640262165/unknown.png

Swan:
Fixes #258

### DIFF
--- a/Quaver.Shared/Database/Maps/OtherGameMapDatabaseCache.cs
+++ b/Quaver.Shared/Database/Maps/OtherGameMapDatabaseCache.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Threading;
+using IniFileParser;
 using Microsoft.Win32;
 using osu.Shared;
 using osu_database_reader.BinaryFiles;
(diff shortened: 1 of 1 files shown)


VB304:
Is there a way to manually point it? I have reinstalled quaver and did a little bit of copy pasting files but nothing seems to work.

PR #2758, opened 2020-11-26
Implement editor layer merging

### LINKED ISSUE #620: Editor: Make merging layers more UX friendly

Illuminati-CRAZ:
Additions/Changes 
- Implement layer merging (Addresses #620)
   - Ctrl+M to merge current layer into the one above
   - Ctrl+Alt+M to merge current layer into the default layer
 - Made EditorLayerRemovedEventArgs and EditorLayerRenamedEventArgs not inherit from EditorLayerCreatedEventArgs
 - Change SelectedLayer.Value to always be DefaultLayer and not null when the defualt layer is selected

 Will need to add the layer merge action to the plugin utility function for creating actions eventually


Swan (changes requested):
There's a few things wrong with this (mostly UI/UX):

1. It took me a long time to figure out how to merge one layer into another. It seems like you have to `CTRL+ALT+A` and then select a layer to merge. Ideally, the UI should provide a way to do all of this for you without keybinds/extra steps.

2. Why can't we merge the default layer into another one? Seems like this would be a good thing to have.

3. The "You cannot merge a layer into itself" notification doesn't give that helpful of information. Similar to the first point, when clicking the merge button in the panel, I would expect to do the merging of two layers there or at the very least have some indication of how to use the feature.

Swan:
Closing since no progress has been made.

Issue #2770, opened 2020-12-03
Add import folder option for other game charts


xDestx:
**Is your feature request related to a problem? Please describe.**
Currently you can import songs from osu! Using import from other games, however StepMania songs all located within specific folders. This means if you want to import all of your SM songs, you have to individually drag each one. This isn't an issue if you had 5-10, but if you have over 1000 it can get quite tedious.

**Describe the solution you'd like**
The normal way to distribute SM packs is by folder, so adding a "import folder" option could be a good addition for players who are migrating from SM (like me). Also, support dragging and dropping a folder onto the game window, just like you would a regular chart.

**Describe alternatives you've considered**
I don't think there are any normal alternatives besides this.

**Additional context**
None!

Jolomann:
Importing packs already works with etterna, so what you could do for the time being is launch etterna once instead of stepmania with your songs and let them import that way

xDestx:
Thanks for the info! I'll do that for the time being. I also wrote something kind of quickly which seems to work here #2771

PR #2771, opened 2020-12-04
Update ImportFile(string file) to allow for directory import

xDestx:
Added one more else if to check if the path is a directory. If a directory, cycle through all files/subdirs and queue all potential map files.

Does not check for replays, skins, databases, etc.

This is my first PR ever yeehaw

### DIFF
--- a/Quaver.Shared/Database/Maps/MapsetImporter.cs
+++ b/Quaver.Shared/Database/Maps/MapsetImporter.cs
@@ -88,23 +88,14 @@ namespace Quaver.Shared.Database.Maps
         }
 
         /// <summary>
-        ///     Tries to import the given file, be it a map, a replay, a skin, etc.
-        ///     <param name="path">path to the file to import</param>
+        ///     Starts importing of queued Maps.
         /// </summary>
-        public static void ImportFile(string path)
+        private static void PostMapQueue()
         {
-            var game = GameBase.Game as QuaverGame;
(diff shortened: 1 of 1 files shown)


xDestx:
for #2770

Quaver.Shared/Database/Maps/MapsetImporter.cs
```diff
@@ -243,6 +243,56 @@ public static void ImportFile(string path)
...
+                        NotificationManager.Show(NotificationLevel.Info, $"Scheduled {Path.GetFileName(subPath)} to be imported!");
+                        Queue.Add(subPath);
+                    }
+                }
+                foreach (var subDir in dirs)
+                {
+                    MapsetImporter.ImportFile(subDir);
+                }
+                // Took this from the first block
+                if (screen.Exiting)
```
IceDynamix, line 265: Maybe extract duplicate code to another function?

xDestx: I was really considering that instead of copying / pasting but wasn't sure if I should or not. It def makes more sense to move it to a function. I'll do it tomorrow and commit again.

xDestx: Okay I moved it to a new function and also added one more which simply checks if a file is of an accepted map type to avoid confusion later on if something were to change.

Swan (approved):

PR #2816, opened 2021-01-18
Fix osu! Directory Path

Kizuna-dev:
Because default path is "Songs" and " dir + "/" " will give the folder relative to Quaver. Now it check if the direction is simply "Songs", if it does, then it give the osu folder path by default, and if not, use the custom dir.


Illuminati-CRAZ:
poggers

PR #2819, opened 2021-01-20
Generate diff bars with slices of strain data instead of hitobjects

IceDynamix:
Groups the strain data from a single difficulty processor into 200 bins instead of creating 200 separate maps and executing 200 processors. Should perform a lot better than before.
Required for https://github.com/Quaver/Quaver.API/pull/105 because the short map nerf eats up all ratings


IceDynamix:
cc @Illuminati-CRAZ
This fixes the things you mentioned

Illuminati-CRAZ:
a section of a map with no hitobjects will have a small white bar associated with it, when it should instead have no bar

PR #2826, opened 2021-01-24
Results screen accuracy graph

IceDynamix:
Adds functionality to the graph dropdown

<details>
    <summary>Example images</summary>
    
    
    
    
    
    
</details>

### DIFF
--- a/Quaver.Shared/Screens/Results/UI/Tabs/Overview/Graphs/ResultsOverviewGraphContainer.cs
+++ b/Quaver.Shared/Screens/Results/UI/Tabs/Overview/Graphs/ResultsOverviewGraphContainer.cs
@@ -8,17 +8,20 @@ using Quaver.Server.Client.Structures;
 using Quaver.Shared.Assets;
 using Quaver.Shared.Config;
 using Quaver.Shared.Database.Maps;
+using Quaver.Shared.Graphics;
 using Quaver.Shared.Graphics.Form.Dropdowns;
 using Quaver.Shared.Helpers;
 using Quaver.Shared.Screens.Results.UI.Tabs.Overview.Graphs.Accuracy;
 using Quaver.Shared.Screens.Results.UI.Tabs.Overview.Graphs.Deviance;
 using Quaver.Shared.Screens.Results.UI.Tabs.Overview.Graphs.Footer;
 using Quaver.Shared.Screens.Selection.UI.FilterPanel.MapInformation.Metadata;
+using Wobble;
 using Wobble.Assets;
 using Wobble.Bindables;
 using Wobble.Graphics;
 using Wobble.Graphics.Sprites;
 using Wobble.Graphics.Sprites.Text;
+using Wobble.Graphics.UI.Buttons;
 using Wobble.Managers;
 
 namespace Quaver.Shared.Screens.Results.UI.Tabs.Overview.Graphs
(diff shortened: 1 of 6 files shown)


YaLTeR:
For the accuracy I had two additional graph ideas in mind which could be interesting (also as in-game displays):
1. Non-decreasing accuracy = accuracy as if all notes from this point onwards were hit with worst possible accuracy.
2. Non-increasing accuracy = accuracy as if all notes from this point onwards were hit with best possible accuracy.

Especially 2. I think could provide clues to where you lost the most acc on the map.

IceDynamix:
Do you mean on a failed or quitted score? While 2. sounds very useful, I'm not sure about how intuitive I can make the graph to convey the idea without creating confusion. Let's hope that very low opacity is enough for this.

Swan:
@IceDynamix Just out of curiosity, why did you opt for this type of graph rather than a simple line graph?

IceDynamix:
The first prototype I made did actually use a "line" graph, where I just plotted each data point to accuracy/time



1. It's hard to make a line graph that looks acceptable, since all the tools I knew to draw with in Wobble were quads (tell me if there is anything else!). Applying some kind of smoothing to the line graph would've been bothersome to implement with what I knew, I already had the idea in my head at this point so I just decided to roll with it
2. I think it fits better with the overall style of the game
3. The rank/accuracy color concept would have been very hard to show in a line graph, here is roughly how it would look like with the same plotting from the prototype

Swan:
Try using [PrimitiveLineBatch](https://github.com/Quaver/Wobble/blob/5efe6424d64d9def1e5283b096559fc199f69fed/Wobble/Graphics/Primitives/PrimitiveLineBatch.cs). It's another type of Drawable where you can specify a list of points, and it will draw a line connecting them. Wobble has a [test scene](https://github.com/Quaver/Wobble/blob/5efe6424d64d9def1e5283b096559fc199f69fed/Wobble.Tests/Screens/Tests/Primitives/TestPrimitivesScreenView.cs) if you need an example on how to use it.

I'm fine with both ways, but I personally would like to see how a line graph would look.

IceDynamix:
Here are lines drawn with the line batch with different thicknesses, I'm not sure what causes those artifacts but it makes this variant very unappealing

Thickness 2
image

Thickness 4
image

Thickness 6
image

IceDynamix:
Here are some results with the graph smoothed out

<details>
<summary>Line graph, thickness 2</summary>
1000 bins

image

500 bins

image

250 bins

image

100 bins

image
</details>

<details>
<summary>Filled graph</summary>
500 bins

image

250 bins

image

100 bins

image
</details>

IceDynamix:
@YaLTeR Here's how the maximum possible accuracy after a fail/quit would look like with 25% opacity
image

Quaver.Shared/Screens/Results/UI/Tabs/Overview/Graphs/ResultsOverviewGraphContainer.cs
```diff
@@ -272,14 +288,62 @@ private void CreateDevianceGraph()
...
+                        const string tooltipText = "Course of accuracy throughout the map.\n" +
+                                                   "If the map was not completed, then it will additionally\n" +
+                                                   "show the accuracy if all subsequent hits had been\n" +
+                                                   "Marvelouses instead.";
+
+                        var game = GameBase.Game as QuaverGame;
+                        tooltipArea.Hovered += (sender, args) => game?.CurrentScreen?.ActivateTooltip(new Tooltip(tooltipText, ColorHelper.HexToColor("#5dc7f9")));
+                        tooltipArea.LeftHover += (sender, args) => game?.CurrentScreen?.DeactivateTooltip();
+
+                        break;
```
IceDynamix, line 342: I will restructure graph classes to be more generic and expandable as a list/array in my next PR once this is merged, to accomodate for the HP graph that's coming next. The current code only a temporary measure, I don't want to add more to this PR with more than it needs to.

YaLTeR:
> @YaLTeR Here's how the maximum possible accuracy after a fail/quit would look like with 25% opacity
> image

This looks like the right implementation, but the idea I had is to draw this for the whole graph. So even in the gameplay area every point is not the accuracy as-is, but accuracy as if all subsequent notes were hit perfectly.

IceDynamix:
I see, but I'm now sure how to calculate the values algorithmically **using the score processor** without being inefficient with iterations

Swan (approved):

PR #2827, opened 2021-01-24
Add retry keybind to Result screen.

FreezePhoenix:


IceDynamix:
1. There are two result screens, one is the one from the alpha ("Result") and one is the current one ("Results"). You made the change on the old screen.
2. (Never thought I'd ever have to say this on a PR but) Duplicate of #2754

PR #2831, opened 2021-01-27
Fix banners not exporting to .qp file

Kizuna-dev:
Change : 
     - Rename old GetBannerPath to GetMapsetBannerPath so we can differentiate what is getting called and for what reason (GetMapsetBannerPath being called to get the mapset banner when loading menu screen, and GetBannerPath for when exporting file)
     - Add a proper GetBannerPath to be able to export the Banner file.
     - Make GetMapsetBannerPath work for etterna chart.
     - Fix Background not being named correctly in the .qua file when exporting an etterna chart from the etterna database. (Because sometime (quite often actually) the file name after the background case is not the same as the background file.) **REQUIRE CHANGE IN Quaver.API** https://github.com/Quaver/Quaver.API/pull/106

### DIFF
--- a/Quaver.Shared/Database/Maps/Mapset.cs
+++ b/Quaver.Shared/Database/Maps/Mapset.cs
@@ -89,10 +89,8 @@ namespace Quaver.Shared.Database.Maps
                                 Logger.Debug($"Successfully converted osu beatmap: {osuPath}", LogType.Runtime);
                                 break;
                             case MapGame.Etterna:
-                                var stepFile = new StepFile(map.Path)
-                                {
-                                    Background = Path.GetFileName(map.BackgroundPath)
-                                };
+                                var stepFile = new StepFile(map.Path);
+
                                 var fileName = StringHelper.FileNameSafeString($"{map.Artist} - {map.Title} [{map.DifficultyName}].qua");
                                 var fileSavePath = $"{tempFolder}/{fileName}";
                                 stepFile.ToQuas().Find(x => x.DifficultyName == map.DifficultyName).Save(fileSavePath);
--- a/Quaver.Shared/Database/Maps/Mapset.cs
+++ b/Quaver.Shared/Database/Maps/Mapset.cs
@@ -89,7 +89,10 @@ namespace Quaver.Shared.Database.Maps
                                 Logger.Debug($"Successfully converted osu beatmap: {osuPath}", LogType.Runtime);
                                 break;
                             case MapGame.Etterna:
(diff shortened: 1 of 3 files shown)


Quaver.Shared/Database/Maps/Mapset.cs
```diff
@@ -89,7 +89,10 @@ public string ExportToZip(bool openInExplorer = true)
                                 Logger.Debug($"Successfully converted osu beatmap: {osuPath}", LogType.Runtime);
                                 break;
                             case MapGame.Etterna:
-                                var stepFile = new StepFile(map.Path);
+                                var stepFile = new StepFile(map.Path)
+                                {
+                                    Background = Path.GetFileName(map.BackgroundPath)
+                                };
```
Swan, line 95: What exactly is this portion intended to be doing? Isn't the background already set during the conversion?

Doing this will also break existing leaderboards for these files when exported and re-imported.

Kizuna-dev: It is, but it don't work for specific reason, please look this video (i will timestamps and explain what's going on) : https://www.youtube.com/watch?v=ogfbq6LCbzg

On current build (0.26.2) : 
0:07 : Background is present in menu.
0:11 : Background is present in game.
0:17 : Export chart.
0:22 : import .qp file.
0:55 : **Background not present in menu after importing the .qp file that got exported.**
0:59 : **Background not in game either.**

On build with this PR :
1:17 : Detect and load DB.
1:56 : Don't have anything to do with this build, osu bg not loading, that's cause i'm using a build were osu.db path didn't get fixed yet.
2:00 : **Banner correctly loaded.**
2:06 : Show the background in game.
2:14 : Export chart.
2:19 : import .qp file.
2:52 : **Banner present even after importing the .qp file.**
2:57 : **Background present in game.**

As to why that "bug" happen in the first place : 
3:15 : background is named bg.
3:22 : background is named differently in the file.

So, what happen is when exporting the .qp file, when creating a "new StepFile" in the code, the background path is set like it's named in the file "aka 2minutesinphotoshop", AND bg is actually imported correctly because of the "copy over background if necessary" just after, which work cause the path is correct there. 

To fix that, i set the background path directly from the mapset.cs, making it sure the background will be correctly named after the background file that will get imported.

Swan: I'm gonna remove this portion for now (the rest of the PR is fine). I need to think of a proper way to fix this. It has other side effects which makes this more complex issue than just the background not being set.

PR #2892, opened 2021-03-12
Add Genre field

Kizuna-dev:
Just added a field for the Genre in the editor, as well as it saving/loading correctly in/from the .qua file.

Tho, i didn't do anything graphics related, as in, the panel will need to be correctly resized by the graphic designer, etc.
Plus i'm not sure if a simple text field like the tags is what is needed, maybe we could have a droplist with a preset of genre ? And we can add/remove as much as we need to.

So yeah, the template is there, i let the graphic designer manage positioning etc.


Illuminati-CRAZ:
What should happen if the music has multiple genres? And what about subgenres?

Kizuna-dev:
> What should happen if the music has multiple genres? And what about subgenres?

Dunno, for the moment it work just like the source/tags, that's why i mentionned "Plus i'm not sure if a simple text field like the tags is what is needed, maybe we could have a droplist with a preset of genre ?

Swan (changes requested):
Two things:

1. Your commit messages need to have better naming. Things like "e" and "pog" are unacceptable and do not allow us to easily search where things have been changed in the commit history.
2. The UI for the metadata editor is broken, and there are fields that overflow outside of the panel. I would suggest increasing the size of the panel.

Kizuna-dev (commented):
As preview, i iprefered resizing field and put them on same row instead of making the panel bigger as it's already is, and i changed the size of the Creator and Difficulty field as they have the most chance to have the smaller text

Swan:
@Kizuna-dev Conflict needs to be fixed before review & merge

Illuminati-CRAZ:
Should change this code to populate genres instead of tags
https://github.com/Quaver/Quaver/blob/b1369ddc62a5c62b5da7806021da20c081732f3f/Quaver.Shared/Screens/Edit/EditScreen.cs#L1333

The change from genre to genres also makes sorting by genre a bit weirder

PR #2898, opened 2021-03-18
Steam integration toggle & Stepmania/Etterna mappack import

tasgon:
Basically what the title says. This lets people disable integration with Steam via the `DISABLE_STEAM` environment variable. It also adds the ability to import mappacks in the form of zip files.

### DIFF
--- a/Quaver.Shared/Screens/Settings/SettingsDialog.cs
+++ b/Quaver.Shared/Screens/Settings/SettingsDialog.cs
@@ -456,6 +456,7 @@ namespace Quaver.Shared.Screens.Settings
                     new SettingsBool(this, "Display Failed Local Scores", ConfigManager.DisplayFailedLocalScores),
                     new SettingsBool(this, "Display Online Friends Notification", ConfigManager.DisplayFriendOnlineNotifications),
                     new SettingsBool(this, "Display Song Request Notifications", ConfigManager.DisplaySongRequestNotifications)
+                    //new SettingsSlider(this, "Number of threads to use for mapset importing", ConfigManager.)
                 })
             };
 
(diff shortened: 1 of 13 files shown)


Illuminati-CRAZ:
It appears there's also something related to showing a grade for a map using local score if the map isn't submitted

tasgon:
Oh, yes, I felt like that might be desirable, as people playing without Steam integration would probably still want to see their local scores in the song listing. (Or at least I would, what do you think?)

Illuminati-CRAZ:
I believe Swan would prefer if there was one pull request per feature added.
(And other things according to [the submission guidelines](https://github.com/Quaver/Quaver/blob/ui-redesign/.github/CONTRIBUTING.md#submitting-a-pull-request) and code style and whatnot)

I agree that using the local score in song listing when global isn't available is a good idea.

Quaver.Shared/Screens/Settings/SettingsDialog.cs
```diff
@@ -456,6 +456,7 @@ private void CreateSections()
                     new SettingsBool(this, "Display Failed Local Scores", ConfigManager.DisplayFailedLocalScores),
                     new SettingsBool(this, "Display Online Friends Notification", ConfigManager.DisplayFriendOnlineNotifications),
                     new SettingsBool(this, "Display Song Request Notifications", ConfigManager.DisplaySongRequestNotifications)
+                    //new SettingsSlider(this, "Number of threads to use for mapset importing", ConfigManager.)
```
IceDynamix, line 459: Probably remove this comment

Swan:
Please split this into multiple pull requests/commits.

PR #2901, opened 2021-03-22
Add health and performance graph

IceDynamix:
# Health graph

Turns red when you fail, useful to detect "fake passes" which are plays with NF that wouldve been a regular pass without NF

dotnet_2021-03-22_20-35-29
dotnet_2021-03-22_20-35-46

# Performance graph

Similar to accuracy graph, colors according to current grade

dotnet_2021-03-22_20-35-16


Swan:
@IceDynamix Maybe for the health graph we could provide the same coloring functionality as the gameplay scoreboard? I think this would make the graph a lot nicer to view.

* 60%-100% - White
* 40%-59% - Yellow
* 1% - 39% - Orange
* 0% - Red
* If NF is on, all subsequent data points after failure are still shown as red.

What do you think?
https://github.com/Quaver/Quaver/blob/ui-redesign/Quaver.Shared/Screens/Gameplay/UI/Scoreboard/ScoreboardUser.cs#L349

IceDynamix:
image

PR #2905, opened 2021-03-25
Use the same resnap algorithm as editor click placement and resnap on paste

IceDynamix:
- The previous, simpler algorithm worked with floats and rounding, so 1ms inaccuracies happened frequently. This one uses the same one as the editor click.
- Resnap on paste! Add shift to paste without resnapping


Quaver.Shared/Screens/Edit/Actions/HitObjects/Resnap/EditorActionResnapHitObjects.cs
```diff
@@ -126,7 +126,7 @@ public void Perform()
                     new EditorActionHitObjectsResnappedEventArgs(Snaps, HitObjectsToResnap));
             }
             else
-                NotificationManager.Show(NotificationLevel.Info, $"No notes to resnap");
+                NotificationManager.Show(NotificationLevel.Info, $"No notes resnapped");
```
Swan, line 129: This notification should be removed when pasting. It's fine to keep it when using the resnap feature though.

Swan (changes requested):

Issue #2913, opened 2021-03-30
Early note hits disappear very high on the playfield, in turn breaking up pattern cohesiveness & reading ability


andrewletz:
**Describe the bug**
In my opinion, this issue is the main reason that many people feel the Quaver engine feels "clunky/bad" when compared to other VSRGs, such as osu!mania or LR2. When hitting notes early, the note disappears entirely, all the way up to a quarter of the playfield. This causes the patterns to break up visually when you hit anything early, which throws off all of your reading and in turn gameplay. This effect is the most notable by far when playing chordjack patterns, as if you accidentally hit some notes early the patterns begin to break up and become more and more confusing.

**To Reproduce**
1. Play any map (preferably dense chordjacks to maximize the effect)
2. Smash your keyboard to get early hits
3. Observe how the notes disappear making the patterns very confusing to read

**Suggested Fix / Expected behavior**
The optimal solution to this problem would be to make early hits "grey out" notes (like when breaking a long note), rather than having them disappear.
A quick solution would be to lower the position at which notes disappear to something like 1/10th of the playfield, rather than 1/4th. This way you can see visually what notes you hit early without _completely_ ruining the cohesiveness of the overall pattern. However, in my opinion, having the notes never disappear would be the best for gameplay.

**Video**
Here is a link to a Youtube video I made showcasing the issue in comparison to osu!mania.
The map is "hurry" at 1.5x speed.
https://www.youtube.com/watch?v=GPWKgYok8a8
You may notice that osu! greys out a note if you hit early once, then will disappear entirely if you hit again, causing a similar issue to Quaver. This can be thought of as a combined version of my two suggested fixes above. However, the position at which they disappear is much lower, making it almost unnoticeable in comparison. I still think that JUST greying out the notes would be optimal, so osu! does not do it perfectly in my opinion.

**Additional context**
In Quaver.Shared/Screens/Gameplay/Rulesets/Keys/HitObjects/HitObjectManagerKeys.cs line 675, you will find "RecycleObjectPosition = (ObjectPositionMagnitude / 4) / ScrollSpeed;". I think this is what causes the notes to disappear at quarter of the playfield's height.

PR #2914, opened 2021-03-30
Option to gray notes if hit early instead of despawning them

IceDynamix:
Addresses #2913 

Difference between old and new (dead color is barely visible because the tint in the default skin is low)
https://user-images.githubusercontent.com/22303902/112993502-9148e800-9169-11eb-9aa9-6d3e7d87c924.mp4

Only grays out notes hit outside of Marv-Great
https://user-images.githubusercontent.com/22303902/112993439-81c99f00-9169-11eb-8290-3c48a86f10cb.mp4

### DIFF
--- a/Quaver.Shared/Config/ConfigManager.cs
+++ b/Quaver.Shared/Config/ConfigManager.cs
@@ -747,6 +747,10 @@ namespace Quaver.Shared.Config
         internal static Bindable<ResultGraphs> ResultGraph { get; private set; }
 
         /// <summary>
+        /// </summary>
+        internal static Bindable<bool> GrayOutEarlyHitNotes { get; private set; }
+
+        /// <summary>
         ///     Dictates whether or not this is the first write of the file for the current game session.
         ///     (Not saved in Config)
         /// </summary>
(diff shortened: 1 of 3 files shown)


andrewletz:
Good point about only graying out early hit notes that are below a certain threshold (i.e Good, Okay, & Miss). In my original issue I was incorrect in stating that it would be best to gray out all early hits. If you do hit a note early and get a Marv/Perf/Great, it should disappear instantly as that visual feedback is important (and expected). This also explains why osu!mania has some notes disappearing and others being grayed out when smashing early hits.

However, it seems osu!mania _only_ grays on early Misses, and will make the note disappear on any early judgement outside of a Miss. I personally think that what IceDynamix has implemented now is a better solution & a good default (gray out early hits below Great), but perhaps it should be made an individual toggle for each judgement so it is up to the user what notes they want to disappear. This would also accomplish what Jolomann requested above.

Something like this (current implementation default is bolded):
"Gray-out rather than disappear early Marvelous judgements" [**Off**/On]
"Gray-out rather than disappear Perfect judgements" [**Off**/On]
"Gray-out rather than disappear Great judgements" [**Off**/On]
"Gray-out rather than disappear Good judgements" [Off/**On**]
"Gray-out rather than disappear Okay judgements" [Off/**On**]
"Gray-out rather than disappear Miss judgements" [Off/**On**]

IceDynamix:
I'll just make a dropdown, using 6 booleans isn't really necessary imo

andrewletz:
Good point with the dropdown, no need to have granularity down to the specific judgement. 
I imagine it would read something like this:
"Gray-out rather than disappear all judgements below and including" [Marvelous/Perfect/Great/**Good**/Okay/Miss]

Swan:
Closing since no progress has been made.

PR #2935, opened 2021-04-21
Handle Config File causing crash

Kizuna-dev:
Simply delete the corrupted config file if an exception is catch, making quaver create a new one.


Illuminati-CRAZ:
Probably would want to have a notification with something to the effect of "hey gamer your config file was corrupted, so we reset it"

YaLTeR:
Here's a cool idea: before deleting try renaming it to `quaver.corrupted.cfg` to avoid potential data loss.

PR #2936, opened 2021-04-21
Fix MP difficulty parsing

YaLTeR:
The previous code failed on difficulty names containing `[something]`. This code isn't perfect either, but it's about the best we can do without sending artist, name and difficulty name as separate strings (which is the correct solution to this, but needs serverside changes, if you could add them please: https://github.com/Swan/Albatross/issues/53).

### DIFF
--- a/Quaver.Shared/Online/MultiplayerGameExtensions.cs
+++ b/Quaver.Shared/Online/MultiplayerGameExtensions.cs
@@ -2,6 +2,7 @@ using System.Linq;
 using System.Text.RegularExpressions;
 using Microsoft.Xna.Framework;
 using Quaver.Server.Common.Objects.Multiplayer;
+using Wobble.Logging;
 
 namespace Quaver.Shared.Online
 {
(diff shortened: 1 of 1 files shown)


Quaver.Shared/Online/MultiplayerGameExtensions.cs
```diff
@@ -21,19 +22,35 @@ public static class MultiplayerGameExtensions
...
+            var spaceBracket = game.Map.IndexOf(" [", dash + 3);
+            if (spaceBracket == -1)
+            {
+                // Also shouldn't happen.
+                Logger.Warning($"Multiplayer map name doesn't contain \" [\" after \" - \": {game.Map}", LogType.Runtime);
+                return "";
             }
 
-            return diffName;
+            return game.Map.Substring(spaceBracket + 2, game.Map.Length - spaceBracket - 3);
```
Swan, line 53: This is something that could be solved by making the regex more robust.

Try: `/.* - .* \[(.*)\]$/`
https://regex101.com/r/Mntx07/1

Does this work?

Swan (changes requested):

Quaver.Shared/Online/MultiplayerGameExtensions.cs
YaLTeR: So the original diffs I noticed the issue on look like this: `artist - title [difficulty [yes]]` and both this regex and some variations I'm trying (`^.+ - .+ \[(.+)\]$`) fail the same way on it. `title [difficulty` becomes the title and `yes]` the difficulty.

That isn't to say the code here doesn't have failure modes (`artist - title [title] [difficulty]` is one of them); really the best solution would be to send the strings separately.

Swan: True. Gonna close this one then & fix server side

PR #2937, opened 2021-04-21
Change new diff name & Fix editor crashing the game because of deleted difficulties

Kizuna-dev:
Change the name of a newly created difficulty to it's creation date only, this fix the map not uploading correctly because of special character.

When deleting a difficulty from a mapset, then try to edit a difficulty from said mapset, the game will crash. This is because of the game is trying to open a non existant file. To prevent this, i added exception handler to prevent the editor from reading those file. I also made it so the game automatically delete the map from the cache, and also reload the database* so the difficulty get removed when going back in the selection screen.

*The reload haven't been tested on a large database, i don't know how much it can impact the editor loading.

(i'm also not sure if this is the best way to handle this situation, but it work.)

### DIFF
--- a/Quaver.Shared/Database/Maps/Map.cs
+++ b/Quaver.Shared/Database/Maps/Map.cs
@@ -383,16 +383,6 @@ namespace Quaver.Shared.Database.Maps
             {
                 case MapGame.Quaver:
                     var quaPath = $"{ConfigManager.SongDirectory}/{Directory}/{Path}";
-
-                    try
-                    {
-                        File.OpenText(quaPath);
-                    } catch (FileNotFoundException)
-                    {
-                        qua = new Qua();
-                        break;
-                    }
-
                     qua = Qua.Parse(quaPath, checkValidity);
                     break;
                 case MapGame.Osu:
--- a/Quaver.Shared/Database/Maps/Map.cs
+++ b/Quaver.Shared/Database/Maps/Map.cs
@@ -383,6 +383,16 @@ namespace Quaver.Shared.Database.Maps
             {
                 case MapGame.Quaver:
                     var quaPath = $"{ConfigManager.SongDirectory}/{Directory}/{Path}";
+
+                    try
+                    {
+                        File.OpenText(quaPath);
+                    } catch (FileNotFoundException)
+                    {
+                        qua = new Qua();
+                        break;
+                    }
+
                     qua = Qua.Parse(quaPath, checkValidity);
                     break;
                 case MapGame.Osu:
--- a/Quaver.Shared/Screens/Edit/UI/AutoMods/EditorAutoModPanelContainer.cs
+++ b/Quaver.Shared/Screens/Edit/UI/AutoMods/EditorAutoModPanelContainer.cs
@@ -1,5 +1,6 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using Microsoft.Xna.Framework;
 using MonoGame.Extended;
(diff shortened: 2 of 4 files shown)


Quaver.Shared/Screens/Edit/UI/AutoMods/EditorAutoModPanelContainer.cs
```diff
@@ -61,7 +62,17 @@ public EditorAutoModPanelContainer(EditScreen screen)
             {
                 if (x == Screen.Map)
                     return;
-
+                try
+                {
+                    File.OpenRead(Directory.GetCurrentDirectory() + "\\Songs\\" + x.Directory + "\\" + x.Path);
```
YaLTeR, line 67: \ is windows-specific, please use /

Quaver.Shared/Screens/Edit/UI/Menu/EditorFileMenuBar.cs
```diff
@@ -347,6 +348,14 @@ private void CreateViewSection()
 
                 foreach (var map in Screen.Map.Mapset.Maps)
                 {
+                    try
+                    {
+                        File.OpenRead(Directory.GetCurrentDirectory() + "\\Songs\\" + map.Directory + "\\" + map.Path);
```
YaLTeR, line 353: same

Quaver.Shared/Screens/Edit/UI/Menu/EditorFileMenuBar.cs
```diff
@@ -347,6 +348,14 @@ private void CreateViewSection()
 
                 foreach (var map in Screen.Map.Mapset.Maps)
                 {
+                    try
+                    {
+                        File.OpenRead(Directory.GetCurrentDirectory() + "/Songs/" + map.Directory + "/" + map.Path);
+                    } catch (FileNotFoundException)
+                    {
+                        break;
+                    }
```
Swan, line 357: This isn't something you're going to want to do every single frame. If the goal is to fix the crash when attempting to load the file, then that should be handled when the user clicks it (and display a notification).

Quaver.Shared/Screens/Edit/UI/AutoMods/EditorAutoModPanelContainer.cs
```diff
@@ -61,7 +62,17 @@ public EditorAutoModPanelContainer(EditScreen screen)
...
-
+                try
+                {
+                    File.OpenRead(Directory.GetCurrentDirectory() + "/Songs/" + x.Directory + "/" + x.Path);
+                } catch (FileNotFoundException)
+                {
+                    mapset.Remove(x.LoadQua());
+                    MapDatabaseCache.RemoveMap(x);
+                    MapDatabaseCache.Load(false);
+                    return;
```
Swan, line 73: I don't think the AutoMod should be responsible for removing files/updating the cache. If files aren't able to be read, then the AutoMod should notify the user of the issue.

Quaver.Shared/Database/Maps/Map.cs
```diff
@@ -383,6 +383,16 @@ public virtual Qua LoadQua(bool checkValidity = false)
...
                     var quaPath = $"{ConfigManager.SongDirectory}/{Directory}/{Path}";
+
+                    try
+                    {
+                        File.OpenText(quaPath);
+                    } catch (FileNotFoundException)
+                    {
+                        qua = new Qua();
+                        break;
+                    }
```
Swan, line 394: Very odd way of handling this. If anything, this error handling should be done when parsing in `Qua.cs`

Swan (changes requested):

Quaver.Shared/Screens/Edit/UI/AutoMods/EditorAutoModPanelContainer.cs
Kizuna-dev: Moved the database update in the EditorScreen file.

Quaver.Shared/Screens/Edit/UI/Menu/EditorFileMenuBar.cs
Kizuna-dev: Found another way and fixed with https://github.com/Quaver/Quaver.API/pull/117

Quaver.Shared/Database/Maps/Map.cs
Kizuna-dev: Done with https://github.com/Quaver/Quaver.API/pull/117

PR #2978, opened 2021-05-06
Check if a config file exist before attempting to read it

Kizuna-dev:
Because of an overlook on my end on my PR https://github.com/Quaver/Quaver/pull/2935 , i forgot to put a check if the file exist before attempting to read it, which mean the game will automatically crash on a fresh installation because no config file exist.

### DIFF
--- a/Quaver.Shared/Config/ConfigManager.cs
+++ b/Quaver.Shared/Config/ConfigManager.cs
@@ -806,32 +806,33 @@ namespace Quaver.Shared.Config
         /// </summary>
         private static void ReadConfigFile()
         {
-            var _configFileDirectory = _gameDirectory + "/quaver.cfg";
-            if (File.Exists(_configFileDirectory))
+            var configFilePath = _gameDirectory + "/quaver.cfg";
+
+            if (File.Exists(configFilePath))
             {
                 try
                 {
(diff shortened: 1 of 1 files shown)


Quaver.Shared/Config/ConfigManager.cs
```diff
@@ -806,16 +806,19 @@ public static void Initialize()
...
+                try
+                {
+                    // Delete the config file if we catch an exception.
+                    var _ = new IniFileParser.IniFileParser(new ConcatenateDuplicatedKeysIniDataParser()).ReadFile(_gameDirectory + "/quaver.cfg")["Config"];
+                }
+                catch (ParsingException)
+                {
+                    Logger.Important("Config file couldn't be read.", LogType.Runtime);
+                    File.Copy(_gameDirectory + "/quaver.cfg", _gameDirectory + "/quaver.corrupted." + TimeHelper.GetUnixTimestampMilliseconds() + ".cfg");
+                    File.Delete(_gameDirectory + "/quaver.cfg");
```
Swan, line 820: `_gameDirectory + "/quaver.cfg"` is repeated multiple times. Use a variable instead.

Swan (changes requested):

PR #2980, opened 2021-05-07
Add checks to ignore menuheader button action if already in that screen

Sceptimis:


MichaelHamilton2 (approved):
The change makes sense would improve notification menu and there is nothing malicious about the change.

Sceptimis:
Oops I should've just done that 😅

PR #3006, opened 2021-05-14
Replays use respective score's judgement windows

Sceptimis:
These changes will make replays use the judgement windows of their respective score, rather than using the selected judgement window preset of the user.


Swan:
The way it currently works is intentional

Sceptimis:
Do you mind elaborating why that is?

Swan:
So players can view replays on different presets - the same way they're able to on the results screen.

Sceptimis:
Maybe it would be better to implement that functionality through the "Convert Score" button on the results screen? It makes more sense to me than having to go and change your global preset? And would still retain the score's original judgement windows in the event where the preset that was used for that replay gets removed?

Swan:
That sounds good

PR #3007, opened 2021-05-15
Add Map and Mapset IPC handlers

IceDynamix:
Enables quaver://map/1234 and quaver://mapset/1234 links to open the map(set) ingame or download it if not present

This PR also adds the /mapset/:id API endpoint

### DIFF
--- a/Quaver.Shared/IPC/QuaverIpcHandler.cs
+++ b/Quaver.Shared/IPC/QuaverIpcHandler.cs
@@ -92,6 +92,13 @@ namespace Quaver.Shared.IPC
         /// <param name="message"></param>
         private static void HandleMapMessage(string message)
         {
+            int mapId;
+            if (!int.TryParse(message, out mapId))
+            {
+                NotificationManager.Show(NotificationLevel.Warning, "The provided ID was not a number!");
+                return;
+            }
+
             var game = (QuaverGame) GameBase.Game;
             if (game.CurrentScreen.Type != QuaverScreenType.Select)
             {
@@ -101,12 +108,6 @@ namespace Quaver.Shared.IPC
             }
 
             var screen = game.CurrentScreen as SelectionScreen;
-            int mapId;
-            if (!int.TryParse(message, out mapId))
-            {
-                NotificationManager.Show(NotificationLevel.Warning, "The provided ID was not a number!");
-                return;
-            }
 
             // Check if we have the map installed
             var map = MapManager.FindMapFromOnlineId(mapId);
@@ -148,6 +149,13 @@ namespace Quaver.Shared.IPC
         /// <param name="message"></param>
         private static void HandleMapsetMessage(string message)
         {
+            int mapsetId;
+            if (!int.TryParse(message, out mapsetId))
+            {
+                NotificationManager.Show(NotificationLevel.Warning, "The provided ID was not a number!");
+                return;
+            }
+
             var game = (QuaverGame) GameBase.Game;
             if (game.CurrentScreen.Type != QuaverScreenType.Select)
             {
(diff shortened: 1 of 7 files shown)


IceDynamix:
Not sure whether a dialog to download a map is necessary so I left it out for now

Quaver.Shared/IPC/QuaverIpcHandler.cs
```diff
@@ -54,7 +62,78 @@ public static void HandleProtocolMessage(string message)
...
+                    NotificationManager.Show(NotificationLevel.Warning,
+                        "You must be in the editor to use this function!");
+            }
+            else if (message.StartsWith("map/"))
+            {
+                var game = (QuaverGame) GameBase.Game;
+                if (game.CurrentScreen.Type != QuaverScreenType.Select)
+                {
+                    NotificationManager.Show(NotificationLevel.Warning,
+                        "You must be in the song select screen to select a map!");
```
YaLTeR: I guess main menu should also work for this (open song select automatically), maybe some other screens? Download?

Quaver.Shared/IPC/QuaverIpcHandler.cs
```diff
@@ -54,7 +62,78 @@ public static void HandleProtocolMessage(string message)
...
+                    return;
+                }
+
+                var screen = game.CurrentScreen as SelectionScreen;
+
+                message = message.Replace("map/", "");
+                int mapId;
+                if (!int.TryParse(message, out mapId))
+                {
+                    NotificationManager.Show(NotificationLevel.Warning, "The provided ID was not a number!");
```
YaLTeR: I'd do parsing checks before the screen check I think so people don't have to quit what they're doing just to find out the message was malformed

Quaver.Shared/IPC/QuaverIpcHandler.cs
```diff
@@ -54,7 +62,78 @@ public static void HandleProtocolMessage(string message)
...
+                    NotificationManager.Show(NotificationLevel.Warning, "You must be logged in to download maps!");
+                    return;
+                }
+
+                // Check if the map exists online
+                var request = new APIRequestMapInformation(mapId);
+                var response = request.ExecuteRequest();
+                if (response.Status == 400 || response.Map?.MapsetId == -1)
+                {
+                    NotificationManager.Show(NotificationLevel.Warning, "The map does not exist!");
```
YaLTeR: Prolly "The map does not exist on the server!" or something like that, should be clearer that it's talking about quaver online maps

Quaver.Shared/IPC/QuaverIpcHandler.cs
```diff
@@ -54,7 +62,78 @@ public static void HandleProtocolMessage(string message)
...
+                    return;
+                } else if (response.Status != 200)
+                {
+                    NotificationManager.Show(NotificationLevel.Error, "Something happened during the request");
```
YaLTeR: image

Quaver.Shared/IPC/QuaverIpcHandler.cs
```diff
@@ -54,7 +62,78 @@ public static void HandleProtocolMessage(string message)
...
+                    var download = MapsetDownloadManager.Download(response.Map.MapsetId, response.Map.Artist, response.Map.Title);
+                    MapsetDownloadManager.OpenOnlineHub();
+
+                    // Auto import
+                    download.Completed.ValueChanged += (sender, args) =>
+                    {
+                        if (game.CurrentScreen.Type == QuaverScreenType.Select)
+                        {
+                            var selectScreen = (SelectionScreen) game.CurrentScreen;
+                            game.CurrentScreen.Exit(() => new ImportingScreen(null, true));
```
YaLTeR: Is there a way to select the correct map after the set has imported? So you don't have to handle the message twice.

IceDynamix: Import already automatically selects the last imported map

YaLTeR: Mapset or map?

IceDynamix: Right... fixed now

Swan:
Going to open two other PRs with some rewrites to your code and also to split the map and mapset handlers since this can be done much simpler & cleaner.

Issue #1606, opened 2020-02-24
Shift+F2 to undo random in song select


Fretzi:
Something like in osu would be great, where when you press Shift+F2, it goes back to the map before you hit F2.
I sometimes accidentaly hit it and don't know what the previous song title was,
or I hit random a couple times, find a map I want, but then hit F2 once too much.
Would be really handy to be able to undo the random in those cases.

PR #3194, opened 2021-08-17
Shift+F2:Undo Random Choice

Renzo904:
-Implements #1606

### DIFF
--- a/Quaver.Shared/Screens/Selection/SelectionScreen.cs
+++ b/Quaver.Shared/Screens/Selection/SelectionScreen.cs
@@ -84,6 +84,11 @@ namespace Quaver.Shared.Screens.Selection
         private Random Rng { get; } = new Random();
 
         /// <summary>
+        ///     Contains the currently history of the random maps
+        /// </summary>
+        public Stack<Map> RngHistory { get; set; } = new Stack<Map>();
+
+        /// <summary>
         ///     Invoked when a random mapset has been selected
         /// </summary>
         public static event EventHandler<RandomMapsetSelectedEventArgs> RandomMapsetSelected;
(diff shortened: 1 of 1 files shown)


Quaver.Shared/Screens/Selection/SelectionScreen.cs
```diff
@@ -574,6 +582,24 @@ private void ChangeScrollSpeed()
...
+            if (AvailableMapsets.Value.Count == 0)
+                return;
+
+            if (RngHistory.Count == 0)
+                return;
+
+            var map = RngHistory.Pop();
+
+            MapManager.Selected.Value = map;
+            RandomMapsetSelected?.Invoke(this, new RandomMapsetSelectedEventArgs(map.Mapset, AvailableMapsets.Value.FindIndex(0, AvailableMapsets.Value.Count, dmap => dmap == map.Mapset)));
```
Swan, line 599: Line is way too long. Either split into two or add a line break.

Quaver.Shared/Screens/Selection/SelectionScreen.cs
```diff
@@ -574,6 +582,24 @@ private void ChangeScrollSpeed()
...
+        {
+            if (AvailableMapsets.Value.Count == 0)
+                return;
+
+            if (RngHistory.Count == 0)
+                return;
+
+            var map = RngHistory.Pop();
+
+            MapManager.Selected.Value = map;
```
Swan, line 598: This needs to handle the case of when the user deletes the previous map in the history and then selects it.

Swan (changes requested):

Quaver.Shared/Screens/Selection/SelectionScreen.cs
Renzo904: Now it will try all the way down to the history until it find an existing one(Fun fact: At the present time, Osu! has this same bug)

Renzo904 (commented):
-Handles the case when a map is not founded(Keeps searching until it finds it)

Issue #3187, opened 2021-08-13
[Options] Allow setting custom fps value in GUI


Jolomann:
**Is your feature request related to a problem? Please describe.**
you have to set a custom fps value in quaver.cfg and many people don't even know it exists. pretty much every time i've mentioned it, the person didn't know the number could be changed

**Describe the solution you'd like**
having it in options would be much more user friendly both in discovery and in practice. it could be a slider and number input field like other options already have

Issue #2995, opened 2021-05-11
[Song Select] Spontaneous insane lag after exiting gameplay


Jolomann:
**Describe the bug**
Seemingly at random, quaver will start to lag to unusable levels (1-3fps) after exiting a song and going back to song select. The game ran fine for hours and then this started to happen, and it now persists on game restarts. I have to wait for about half a minute before the lag goes away

I'm not sure what could be causing it but here are some observations

1) The lag happens regardless of search filters and keywords, so it's probably not search related
2) If you manage to either go to main menu or enter gameplay, the lag disappears instantly
3) High process priority made no difference
4) For no obvious reason, the same lag will sometimes happen after alt-tabbing and tabbing back into the game, but it only happened to me once
5) Quaver doesn't use more system resources while lagging like described

**To Reproduce**
1. Get bad quaver RNG (Really not sure what the root cause is)
2. Go into gameplay
3. Exit to song select
4. Admire powerpoint presentation

**Expected behavior**
More frames

**Version:**
 - Quaver v0.28.0

Illuminati-CRAZ:
Lag might be caused by gameplay preview. I've found that the amount of lag changes if you change where in the map you're looking at. For me it also seems to occur more often after switching to unlimited fps.

IceDynamix:
bumping this again, because this is by far the biggest issue i have with the game as of now. every session usually has about 2-3 occurances of the bug i now call the "2 fps bug" and i either have to
- wait until it resolves by itself after several minutes
- spam escape until it resolves by itself so it jumps to main menu which restores the original fps (can also take minutes)
- or close the process

fixing this should be the #1 priority imo 

occurances on stream
- https://www.twitch.tv/videos/1434951937
- https://www.twitch.tv/videos/1434951936

PR #3199, opened 2021-08-18
Add custom fps slider in the gui

Kizuna-dev:
Add a slider to change custom fps limit without changing it manually in the .cfg file.

Also limited the custom fps to 1000fps otherwise it be a nightmare to select the fps you want with the slider.

https://github.com/Quaver/Quaver/issues/3187


Swan (changes requested):
For some reason this causes an extreme FPS drop no matter what I have the value at. Possibly related to #2995 because they are known to use a custom FPS limit.

image

PR #3201, opened 2021-08-18
Remove bpm limit when creating timing lines for gameplay

### LINKED ISSUE #2982: Remove BPM limit when generating timing lines
I would like to pack some timing lines within less than 6.00060006 ms of each other

Remove [lines 102-104](https://github.com/Quaver/Quaver/blob/9016b5ebe26c35d80cade60415f9bbeb33d2d49f/Quaver.Shared/Screens/Gameplay/Rulesets/Keys/Playfield/Lines/TimingLineManager.cs#L102-L104) from the TimingLineManager and then remove the Math.Min() in line 105

Place a new timing point every measure, which causes a considerable increase in file size in some cases
This alternative also doesn't work if the lines need to be within 1 ms of each other

**Additional context**
Removing the limit might also make evening's aleph-0 map properly viewable lol

Illuminati-CRAZ:
resolves #2982

Was worried that this would break some maps, so I wrote a thing to find maps in my songs folder with timing points that have >9999 bpm and that didn't have their lines hidden.

Generally for the maps I checked, they were either used for teleporting (when BPM affects SV) or for overriding beat snap colors, both of which this change didn't appear to negatively impact.

The maps I found that were negatively impacted were:
[Helblinde - The Solace of Oblivion [A Lie's Asgard]](https://osu.ppy.sh/beatmapsets/596105#mania/1274057): lags like crazy
[LeaF - Aleph-0 (extended ver.) [Boundless]](https://osu.ppy.sh/beatmapsets/1226110#mania/2549754): either doesn't start or takes a really really long time to start
Utata-P feat. Yuzuki Yukari - Shiawase ni Nareru Kakushi Command ga Arurashii [Cheat Code]: laggier

I did not find any ranked maps that were broken by this change.

### DIFF
--- a/Quaver.Shared/Screens/Gameplay/Rulesets/Keys/Playfield/Lines/TimingLineManager.cs
+++ b/Quaver.Shared/Screens/Gameplay/Rulesets/Keys/Playfield/Lines/TimingLineManager.cs
@@ -99,10 +99,7 @@ namespace Quaver.Shared.Screens.Gameplay.Rulesets.Keys.Playfield.Lines
 
                 var signature = (int)map.TimingPoints[i].Signature;
 
-                // Max possible sane value for timing lines
-                const float maxBpm = 9999f;
-
-                var msPerBeat = 60000 / Math.Min(Math.Abs(map.TimingPoints[i].Bpm), maxBpm);
+                var msPerBeat = 60000 / Math.Abs(map.TimingPoints[i].Bpm);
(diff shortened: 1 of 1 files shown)


Swan:
The whole purpose of adding the limit in the first place was to prevent the immense amount of lag that some maps have from disrupting screens like song select and multiplayer. Unless this directly provides a fix for that, then it won't be merged.

PR #3212, opened 2021-08-22
Use tint depending on status for FFA/BR multiplayer player card

Adrriii:
This aims at making it more visible when you are host of the game (primarily for host rotation lobbies)
Then also improves visibility of player statuses (idle/ready/missing map)
This is only for FFA and Battle Royale, since Team vs mode uses this as a team indicator.

Quaver_sByiBqL1Mv
Quaver_scjbnDy2Ra
Quaver_9qzFTll4mm


Adrriii:
One thing : I noticed some constants (especially colors) are hardcoded, might be a good thing if i improve this in this PR ?

PR #3231, opened 2021-09-07
Add skinning for the results screen

Kizuna-dev:
This will add basic skinning (simply changing images) for each element in the results screen (outside of number/judgements).

This also add ResultsBackgroundType which allow to change the header, the skinner can choose between "Header","Background","None".
Header will act as the current way. (this is also the default value)
Background will take the whole screen, so it will be the map background instead of the results screen background.
None is none.

This also make use of the background filter with the ResultsBackgroundFilterAlpha config.

Also not sure if this was intended, but the result screen background use different X and Y position, which mean there were a visible grey line on the right, as it's probably not wanted, i changed the value to 0.

### DIFF
--- a/Quaver.Shared/Skinning/Menus/SkinMenuResults.cs
+++ b/Quaver.Shared/Skinning/Menus/SkinMenuResults.cs
@@ -2,6 +2,7 @@ using IniFileParser.Model;
 using Microsoft.Xna.Framework;
 using Microsoft.Xna.Framework.Graphics;
 using Quaver.Shared.Config;
+using Quaver.Shared.Screens.Results.UI;
 
 namespace Quaver.Shared.Skinning.Menus
 {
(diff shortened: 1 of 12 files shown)


Quaver.Shared/Skinning/Menus/SkinMenuResults.cs
```diff
@@ -0,0 +1,68 @@
...
+        public Texture2D ResultsLabelTotalScore { get; private set; }
+        public Texture2D ResultsLabelScore { get; private set; }
+        public Texture2D ResultsLabelBlueTeam { get; private set; }
+        public Texture2D ResultsLabelRedTeam { get; private set; }
+        public Texture2D ResultsMultiplayerTeamPanel { get; private set; }
+        public Texture2D ResultsScoreContainerPanel { get; private set; }
+        public Texture2D ResultsGraphContainerPanel { get; private set; }
+        public Texture2D ResultsMultiplayerFFAPanel { get; private set; }
+        public Texture2D ResultsBackground { get; private set; }
+        public string ResultsBackgroundType { get; private set; }
```
Swan, line 27: You're going to want to use an enum here rather than a string. [ConfigHelper.ReadEnum()](https://github.com/Quaver/Quaver/blob/ui-redesign/Quaver.Shared/Config/ConfigHelper.cs#L271) will help with parsing.

Swan (changes requested):

Swan (changes requested):
There's a regression with the background header. The default skin's background is lighter than what it used to be.

Old




New

Swan (approved):

PR #3242, opened 2021-09-20
Add unranked modifier leaderboard support

Swan:
- Scores using the NLN, FLN, or INV modifiers are now able to be submitted to the server.
- Migrated to the new scoreboard API endpoint.
- Added the "All" leaderboard.
- Difficulty ratings for NLN, FLN, and INV are now calculated on the fly during gameplay and the results screen.
- Performance ratings for scores with NLN, FLN, INV now use the correct difficulty rating.


IceDynamix:
nsv not submittable, tragic

PR #3290, opened 2021-10-30
implement steam's rich presence.

surepy:
This PR aims to implement Steam's [Enhanced Rich Presence](https://partner.steamgames.com/doc/features/enhancedrichpresence) onto Quaver. 

which adds the benefit of friend grouping if in the same multiplayer lobby, and not having to respond "what are you playing" in steam messages. ~~not really, people message anyways.~~

~~important note: This needs separate testing as I cannot test it because I don't have access to Steamworks. lol. that's why this is a draft..~~ see below
this is the localization file that I would like to commit.

```
"lang"
{
	"english"
	{
		"tokens"
		{
			"#State"	"%state%"
			"#Details"	"%details%"
			"#Status"	"{#State}: {#Details}"
		}
	}
}
```

### DIFF
--- a/Quaver.Shared/Screens/Downloading/DownloadingScreen.cs
+++ b/Quaver.Shared/Screens/Downloading/DownloadingScreen.cs
@@ -853,6 +853,9 @@ namespace Quaver.Shared.Screens.Downloading
         {
             try
             {
+                SteamManager.SetRichPresence("State", "In the menus");
+                SteamManager.SetRichPresence("Details", "Downloading Maps");
+
                 DiscordHelper.Presence.Details = "Downloading Maps";
                 DiscordHelper.Presence.State = "In the menus";
                 DiscordHelper.Presence.LargeImageText = OnlineManager.GetRichPresenceLargeKeyText(ConfigManager.SelectedGameMode.Value);
--- a/Quaver.Shared/Screens/Downloading/DownloadingScreen.cs
+++ b/Quaver.Shared/Screens/Downloading/DownloadingScreen.cs
@@ -853,15 +853,11 @@ namespace Quaver.Shared.Screens.Downloading
         {
             try
             {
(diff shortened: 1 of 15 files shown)


Swan:
@surepy What exactly is preventing you from testing? You should have access to the majority of the features in the API already.

Quaver.Shared/Screens/Downloading/DownloadingScreen.cs
```diff
@@ -853,6 +853,9 @@ private void SetRichPresence()
         {
             try
             {
+                SteamManager.SetRichPresence("State", "In the menus");
+                SteamManager.SetRichPresence("Details", "Downloading Maps");
+
                 DiscordHelper.Presence.Details = "Downloading Maps";
                 DiscordHelper.Presence.State = "In the menus";
                 DiscordHelper.Presence.LargeImageText = OnlineManager.GetRichPresenceLargeKeyText(ConfigManager.SelectedGameMode.Value);
```
Swan, line 861: It'd be better if we can change both Discord and Steam in one method, so we can avoid having to do this twice every single time.

Swan (changes requested):

Quaver.Shared/Screens/Downloading/DownloadingScreen.cs
surepy: fair enough, I'll do that.

surepy:
> @surepy What exactly is preventing you from testing? You should have access to the majority of the features in the API already.

the Rich Presence Tester didn't work the day I made this apparently now it just decides to ¯\_(ツ)_/¯

image

Swan:
@surepy I've uploaded the file you provided, but there's still errors being thrown in the tester.

image

surepy:
lmao i forgot a `#`

Swan (approved):

PR #3297, opened 2021-11-02
Add gamepad support for gameplay inputs

YaLTeR:
Controller players rejoice.

Requires https://github.com/Quaver/Wobble/pull/116

This PR contains two changes:
1. XnaKeyHelper is moving to Wobble.Helpers, so it's deleted from here and a few imports are added.
2. Several places which used to deal with Keys now deal with GenericKey (see the Wobble PR).

The net result is that it's now possible to bind gamepad buttons to gameplay inputs in addition to keyboard keys. The configuration format for keyboard keys is preserved, while gamepad keys are written with the GP_ prefix, consistent with how they are displayed in the options.

This PR changes just the gameplay inputs leaving all other inputs as is for now.

I did a basic test on my Linux system with an Xbox 360 for Windows Wireless controller and it seemed to work. Would be nice if someone on Windows with an actual IIDX-like controller could give it a go. I've heard they are recognized as PlayStation 3 controllers or something like that.

EDIT: Hubert tried and the game didn't register the inputs for him, maybe there's something extra you need to do on Windows? Idk

EDIT 2: Sceptimis separately confirmed it works on his Linux setup with xbox one and ps3 gamepads

EDIT 3: AiAe's switch pro controller worked on Windows

EDIT 4: Also worked for me with my controller on Windows.


Sceptimis:
What controller/gamepad was Hubert using?

Sceptimis:
As mentioned, Xbox One and PS3 Dualshock controllers worked out-of-the-box for me, albeit through a USB/Wired connection. I just tested a Bluetooth game-pad (PowerA GameCube Controller for Nintendo Switch), which after pairing and starting Quaver with the patch applied, did not work. Unfortunately, my Xbox One controller is not the Bluetooth-capable version, so I couldn't verify functionality for that controller, but I suspect it wouldn't work either.

YaLTeR:
> What controller/gamepad was Hubert using?

He tried with an IIDX controller pretending to be a PS3 controller, and with a "new xbox controller".

YaLTeR:
AiAe has a Switch Pro controller which worked on Windows with these changes.

YaLTeR:
Also worked for me on Windows with my Xbox 360 for Windows controller.

YaLTeR:
AiAe got it to work on macOS using Steam Big Picture controller settings.

YaLTeR:
cat got his iidx controller to work only with a bit of weird Steam Big Picture setup, however he says that after getting it to work, he doesn't encounter issues that he had with joy2key, so this is an improvement after it's set up.

I'd say this PR is good to go then; we can figure out how to make it work with more controllers out of the box afterwards.

YaLTeR:
I remade the Wobble PR to use the raw joystick inputs rather than gamepad (which relies on the controller DB). This made the iidx controller (and probably all others) work out of the box at the cost of nice button names (so now it's GP_1, GP_2 instead of, say, GP_A, GP_X). However, it seems that both osu! and LR2 also use these raw button names so it's all good.

cat confirmed it works for him now without the Big Picture binding.

Swan:
For blog post

---

- Fixed a bug where chat messages sometimes overlapped each other.

# Controller Support

You can now directly bind controller buttons to gameplay inputs and pause and skip intro keys. This includes joysticks, gamepads and custom rhythm game controllers. It works across all operating systems and provides much better input accuracy compared to JoyToKey and similar solutions.

Swan (approved):

PR #3302, opened 2021-11-08
ChatMessageScrollContainer: Lock MessageHistoryQueue

YaLTeR:
It was being locked from RunRequestHistoryTask() but not from FlushMessageHistoryQueue(). I think this should fix overlapping messages in chat.


Swan:
for blog post

---

- Fixed a bug where chat messages sometimes overlapped each other.

PR #3330, opened 2021-12-13
Remove hitobjects from selection when removed by plugin or livemapping

### LINKED ISSUE #2631: Selected HitObjects removed by plugin stay in selection
To reproduce:
1. Select some objects
2. actions.RemoveHitObjectBatch(state.SelectedHitObjects)
3. Ctrl + C to copy HitObjects still in selection
4. Ctrl + V to restore HitObjects previously removed

Problem does not occur when removing objects by right-clicking or by pressing the Delete key.

Illuminati-CRAZ:
fixes #2631

might make more sense to have the editor actions themselves remove from selection instead of manually adding it to every instance of removal

### DIFF
--- a/Quaver.Shared/Screens/Edit/Actions/EditorActionManager.cs
+++ b/Quaver.Shared/Screens/Edit/Actions/EditorActionManager.cs
@@ -55,7 +55,7 @@ namespace Quaver.Shared.Screens.Edit.Actions
     {
         /// <summary>
         /// </summary>
-        private EditScreen EditScreen { get; }
+        public EditScreen EditScreen { get; }
 
         /// <summary>
         /// </summary>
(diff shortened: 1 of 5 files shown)


YaLTeR:
> might make more sense to have the editor actions themselves remove from selection instead of manually adding it to every instance of removal

Would that also make it so Ctrl-Z selects them back? I think that's a good idea.

Illuminati-CRAZ:
currently none of the ways of removing restore selection on undo

YaLTeR:
If actions did it would that make all of them restore selection?

Illuminati-CRAZ:
the undo method of the action could be coded so that the selection is restored, and then all of them would restore selection

YaLTeR:
That would be cool yeah. Would that be a lot of work?

Illuminati-CRAZ:
Shouldn't be, I think a different action already restores selection on undo actually

YaLTeR:
Alright, could you do that then?

Illuminati-CRAZ:
yeah sure

Illuminati-CRAZ:
now that i think about it undoing the deselection on the notes is kinda weird

im thinking you could have like 5 notes selected, a few of them are removed, the rest are deselected, and then the remove action in undone

you end up with a few notes selected out of the original selection, and the new selection is probably not what you want

or in some other situation you unexpectedly have a note selected from undoing an earlier removal, and that interferes with copy-pasting or some plugin functionality

YaLTeR:
What if undoing the deselection would add notes to current selection instead of replacing it?

Illuminati-CRAZ:
same potential problems

YaLTeR (approved):

Issue #3323, opened 2021-12-11
There should be ignore list for when exporting & uploading mapsets for hidden OS files


AiAe:
`.DS_Store` - MacOS
`Thumbs.db` - Windows

PR #3336, opened 2021-12-19
Mapset: Export only allowed file extensions

YaLTeR:
Fixes multiplayer uploads failing for some osu! mapsets (the ones containing .osb storyboards for example) because the server denies files that don't have an allowed extension of MIME type.


AiAe (approved):

AiAe:
Also fixes https://github.com/Quaver/Quaver/issues/3323

PR #3420, opened 2022-03-12
Make the avatar use the new mask script for gradient

Kizuna-dev:
Make use of https://github.com/Quaver/Wobble/pull/120 so that the mask can have different alpha value

Does require https://github.com/Quaver/Wobble/pull/120

Here is a result of the effect 
Quaver_tdwKVPF3Vj

using this mask
avatar-mask

### DIFF
--- a/Quaver.Shared/Screens/Results/UI/Header/Contents/ResultsScreenHeaderAvatar.cs
+++ b/Quaver.Shared/Screens/Results/UI/Header/Contents/ResultsScreenHeaderAvatar.cs
@@ -62,7 +62,8 @@ namespace Quaver.Shared.Screens.Results.UI.Header.Contents
 
             GameBase.Game.ScheduledRenderTargetDraws.Add(() =>
             {
-                Avatar.Image = Avatar.PerformBlend(Avatar.Image, SkinManager.Skin?.Results?.ResultsAvatarMask ?? UserInterface.ResultsAvatarMask);
+                var mask = SkinManager.Skin?.Results?.ResultsAvatarMask ?? UserInterface.ResultsAvatarMask;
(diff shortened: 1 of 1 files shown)


Quaver.Shared/Screens/Results/UI/Header/Contents/ResultsScreenHeaderAvatar.cs
```diff
@@ -50,13 +52,18 @@ public ResultsScreenHeaderAvatar(float size, Bindable<ScoreProcessor> processor)
...
+                        Avatar.Image = SteamManager.UserAvatars[steamId];
                     else
-                        Avatar.AvatarSprite.Image = UserInterface.UnknownAvatar;
+                        Avatar.Image = UserInterface.UnknownAvatar;
                 }
             }
+
+            GameBase.Game.ScheduledRenderTargetDraws.Add(() =>
+            {
+                Avatar.Image = SpriteAlphaMaskBlend.PerformBlend(Avatar.Image, SkinManager.Skin?.Results?.ResultsAvatarMask ?? UserInterface.ResultsAvatarMask);
```
Swan, line 65: Split into two lines. It's too long.
```cs
var mask = SkinManager.Skin?.Results?.ResultsAvatarMask ?? UserInterface.ResultsAvatarMask;
Avatar.Image = SpriteAlphaMaskBlend.PerformBlend(Avatar.Image, mask);
```
Also, see my comment [here](https://github.com/Quaver/Wobble/pull/120#discussion_r825487367). Memory leak needs to be fixed. Either by caching the avatar or disposing it in `Destroy()`. The latter would probably be a better option if people are trying to modify their skin.

Swan (changes requested):

PR #3446, opened 2022-04-08
Add leaderboard toggle button

4techguns:
In this pull request, a new, small feature is added to the song selection screen. It is a tiny tab that, when clicked, toggles the leaderboard's visibility. This can be especially useful for those who do not want a leaderboard on screen.

The leaderboard is also destroyed when hidden, and re-instantiated when shown again, so if the player is having lag issues on a lower-end machine due to the leaderboard, they can just simply hide it.

Here is the feature in-game:

https://user-images.githubusercontent.com/42378704/162434774-fdcf9133-a597-40b8-b441-900826226bbd.mp4

### DIFF
--- a/Quaver.Shared/Screens/Selection/SelectionScreenView.cs
+++ b/Quaver.Shared/Screens/Selection/SelectionScreenView.cs
@@ -10,6 +10,7 @@ using Quaver.Shared.Database.Profiles;
 using Quaver.Shared.Graphics.Menu.Border;
 using Quaver.Shared.Helpers;
 using Quaver.Shared.Scheduling;
+using Quaver.Shared.Screens.Menu.UI.Jukebox;
 using Quaver.Shared.Screens.Menu.UI.Visualizer;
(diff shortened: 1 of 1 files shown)


IceDynamix:
having the big gaping empty space does not seem productive and looks out of line with the rest of the game, so if it's about not seeing a leaderboard, why not just switch to one of the other useful panels, such as the gameplay preview panel?

the leaderboard should be destroyed after it left the screen with a movement transition, just like how it enters but in reverse. it feels weird seeing a floating button traveling without cause because the leaderboard disappears too quickly

skinning has also not been considered, so the button is going to look even more out of line with skins that customize the song select and leaderboard

and finally, if you're having lag... it's definitely not because of the leaderboard

it feels like you're trying to solve a problem that doesn't exist with a solution that doesn't add anything of value

4techguns:
You have a fair point, and I didn't really put much thought into this feature when making. Thanks for your input on this! Closing

PR #3499, opened 2022-06-10
Don't scale miss window for harder default judgements

Zexenuxx:
Miss window for the default presets have been updated. 
If we were to scale down the miss judgement on harder judgements, it would make it easier to acc because
every miss judgement gets smaller and smaller, while miss judgements on easier timing windows get bigger and bigger.

To elaborate, let's say two players, one using peaceful, the other using impossible, both hit a note at 150ms.
The player with peaceful judgement receives an "Ok", while the player with impossible judgement receives a miss.
The player with the peaceful judgement takes a stronger blow to their accuracy than the one with impossible judgement,

### DIFF
--- a/Quaver.Shared/Database/Judgements/JudgementWindowsDatabaseCache.cs
+++ b/Quaver.Shared/Database/Judgements/JudgementWindowsDatabaseCache.cs
@@ -115,7 +115,10 @@ namespace Quaver.Shared.Database.Judgements
             windows.Great /= (float) Math.Pow(constant, n);
             windows.Good /= (float) Math.Pow(constant, n);
             windows.Okay /= (float) Math.Pow(constant, n);
-            windows.Miss /= (float) Math.Pow(constant, n);
+            // Only scale the miss timing on easier windows because scaling the bad timings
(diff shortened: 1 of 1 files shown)


IceDynamix:
different wording for the reasoning behind this pr

scaling the miss window on a harder judgement setting decreases the miss window size, which can lead to greater accuracy when compared to an easier judgement setting. this happens because hit that are outside the miss window on the harder setting but inside the miss window on the easier setting will penalize the user on the easier setting and not the hard one.

the recommended advice to prevent this issue has been to create a copy of the existing judgement settings with the miss window edited to be the same as the standard miss window. this PR goes ahead of this workaround, so players do not have to create a custom judgement setting anymore.

however, this would also cause the bad window to become larger than the miss window on easier judgements, hence why the scaling is only applied to easier judgements and not harder judgements.

IceDynamix:
clarification that this change only affects the harder preset judgements, not the standard one

Swan:
@Zexenuxx @IceDynamix What about doing this for the Okay window as well?

PR #3527, opened 2022-06-27
Rewrite LoadPlaylists to use Tasks

AiAe:
LoadPlaylists now uses Tasks and its rewriten in better optimized way.

Changed Dictionary to ConcurrentDictionary because its thread safe.
Tasks uses thread pool.

Some tests:
From original code: RUNTIME - DEBUG: 00:00:37.1226258
New code: RUNTIME - DEBUG: 00:00:07.1438216

All tests are done with https://www.mediafire.com/file/cb6yrh0f0bpyyuv/quaver.zip/file

### DIFF
--- a/Quaver.Shared/Database/Playlists/PlaylistManager.cs
+++ b/Quaver.Shared/Database/Playlists/PlaylistManager.cs
@@ -29,11 +29,6 @@ namespace Quaver.Shared.Database.Playlists
         public static List<Playlist> Playlists { get; private set; } = new List<Playlist>();
 
         /// <summary>
-        ///     Tasks list
-        /// </summary>
-        public static List<Task> Tasks { get; private set; } = new List<Task>();
-
-        /// <summary>
         ///     The currently selected playlist
         /// </summary>
         public static Bindable<Playlist> Selected { get; } = new Bindable<Playlist>(null);
@@ -146,6 +141,7 @@ namespace Quaver.Shared.Database.Playlists
         /// </summary>
         private static void LoadPlaylists()
         {
+            var tasks = new List<Task>();
             Playlists = new List<Playlist>();
             Selected.Value = null;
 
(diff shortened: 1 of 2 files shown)


Quaver.Shared/Database/Playlists/PlaylistManager.cs
```diff
@@ -25,6 +28,11 @@ public static class PlaylistManager
         /// </summary>
         public static List<Playlist> Playlists { get; private set; } = new List<Playlist>();
 
+        /// <summary>
+        ///     Tasks list
+        /// </summary>
+        public static List<Task> Tasks { get; private set; } = new List<Task>();
+
```
Swan, line 35: This shouldn't be a property if it's only being used in one method and nowhere else.

Quaver.Shared/Database/Playlists/PlaylistManager.cs
```diff
@@ -133,6 +141,16 @@ private static void CreateTables()
...
 
+        /// <summary>
+        ///     Convert list to concurrent dictionary
+        /// </summary>
+        public static ConcurrentDictionary<TKey, TValue> ToConcurrentDictionary<TKey, TValue>
+            (this IEnumerable<TValue> source, Func<TValue, TKey> valueSelector)
+        {
+            return new ConcurrentDictionary<TKey, TValue>
+                (source.ToDictionary(valueSelector));
+        }
```
Swan, line 152: Move this extension method into its own helper class/file - `ConcurrentDictionaryExtensions.cs`

The signature should also be on one line instead of two.

Quaver.Shared/Database/Playlists/PlaylistManager.cs
```diff
@@ -146,32 +164,34 @@ private static void LoadPlaylists()
...
 
+                Task.WaitAll(Tasks.ToArray());
+
                 Playlists = Playlists.Concat(playlists).ToList();
 
                 foreach (var playlist in playlists)
-                    Logger.Important($"Loaded Quaver playlist: {playlist.Name ?? ""} w/ {playlist.Maps?.Count ?? 0} maps!", LogType.Runtime);
+                    Logger.Important(
+                        $"Loaded Quaver playlist: {playlist.Name ?? ""} w/ {playlist.Maps?.Count ?? 0} maps!",
+                        LogType.Runtime);
```
Swan, line 194: Formatting.

Swan (changes requested):

Swan (approved):

Issue #3386, opened 2022-02-06
Top right of screen will say "USER ONLINE" when attempting to view online users while offline


Illuminati-CRAZ:
**Describe the bug**
While the user if offline, if they press F8 or F9 and try to look at online users, the UI will unexpectedly say "USER ONLINE" in the top right. This is usually used to say how many people are online when the player is online, but as a funny coincidence it says "USER ONLINE" when offline.

**To Reproduce**
1. Open Quaver while offline (being online and logging out appears to not work)
2. Press F8 or F9
3. Observe the UI saying "USER ONLINE" in the top right

**Expected behavior**
The UI does not say that the user is online when the user is not online

**Screenshots
![252022 22-58-3-337](https://user-images.githubusercontent.com/42056854/152670922-80b3004f-af3b-4894-9cbc-86ed706334fd.jpg)
**

Tested on version 0.31.0

PR #3537, opened 2022-06-28
Display OFFLINE if client is disconnected

AiAe:
Fixes https://github.com/Quaver/Quaver/issues/3386


Swan (changes requested):
I don't think this will handle the case of when you are online and then disconnect. Calling `SetUserCount()` in `OnConnectionStatusChanged()` should fix this.

Issue #3543, opened 2022-06-30
Skins dropdown makes buttons not working


AiAe:
**To Reproduce**
Go to select screen
Open Options
Go to Skin and open Custom Skin dropdown (maybe scroll in it?) and close it
Go back to Video and toggles wont be clickable/hovarable (it also effects other screens until you change one with keyboard)

**Desktop (please complete the following information):**
 - OS: macOS
 - Version 0.31.4

Dropdowns could be the reason for when in other places of the game everything becomes unclickable?

AiAe:
Did some more testing, I think having some skins is required.
And the game to be fullscreen but windowed?

PR #3562, opened 2022-07-10
Dropdown fixes

AiAe:
ToDos:

- [x] Opening options in DropdownItem close() is called 31 times
Fixed by moving ClickedOutside to Dropdown, it resolves the issue.

- [x] Dropdown destroy is called only for the visible dropdowns
Presented in Options when opening the dialog its creating all dropdowns, but on closing it only closes the visible dropdowns, which leaves the rest in memory and might be the reason why its blocking everything on screen to be clickable.
Possible fix: to create only the visible dropdowns in the selected tab or to destroy everything when the dialog is closed

- [x] Dropdown close is triggered for all dropdowns instead just for the one that we interacted with

Debugging info: 
Starting the game and going to Select screen dropdowns items count is just 30.
Starting the game going to Options closing and going to Select screen dropdowns items count is 102.

This PR will try to resolve all issues in https://github.com/Quaver/Quaver/issues/3543

### DIFF
--- a/Quaver.Shared/Graphics/Form/Dropdowns/Dropdown.cs
+++ b/Quaver.Shared/Graphics/Form/Dropdowns/Dropdown.cs
@@ -364,7 +364,8 @@ namespace Quaver.Shared.Graphics.Form.Dropdowns
             if (ItemContainer.ScreenRectangle.Contains(mousePoint) || ScreenRectangle.Contains(mousePoint))
                 return;
 
-            Close();
+            if(Opened)
+                Close();
         }
     }
 }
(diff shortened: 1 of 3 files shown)


Quaver.Shared/Graphics/Form/Dropdowns/Dropdown.cs
```diff
@@ -135,13 +137,15 @@ public class Dropdown : ImageButton
...
             Clicked += OnClicked;
+            ClickedOutside += OnClickedOutside;
         }
 
         /// <inheritdoc />
         /// <summary>
         /// </summary>
         public override void Destroy()
         {
+            Logger.Debug("Destroy dropdown!!!!!!!!!", LogType.Runtime);
```
IceDynamix, line 148: destroy dropdown!!!!!!!!¡¡¡

AiAe:
Im unable to fix anything for the other two issues, if someone else can take them.

AiAe:
Managed to fix everything. 
Only found new bug with them in the Options, if you open dropdown and switch the tab and go back to the previous one, even that its closed its still presented and anything that was under it is not clickable.

AiAe:
Managed to fix everything that I aimed for. 
Only discovered spam opening the dropdowns still causes the issue in Options, but when closed and reopen is fine.

Issue #2131, opened 2020-06-13
Timing lines disappear in gameplay preview if seeking backwards in time


Illuminati-CRAZ:
Occurs with song select map preview and editor gameplay preview

PR #3615, opened 2022-08-24
Fix timing lines not appearing after seeking backwards in gameplay

Illuminati-CRAZ:
fixes https://github.com/Quaver/Quaver/issues/2131

tbh I don't really know what I'm doing here, but it seems to work and there doesn't seem to be any significant drops in performance (except for [cherry blossoms explode across the dying horizon](https://quavergame.com/mapset/map/35926), but that has like 10k timing lines and even then it performs okay-ish (not really))


IceDynamix:
@YaLTeR i dont have enough experience to comment on this regarding performance and what happens under the hood, can you take a look at this?

YaLTeR:
Uhh, I'll need to take a deeper look if I want to understand this change

Swan (approved):
Seems fine. This is similar to how I would've done it and how seeking backwards works with other objects

PR #3666, opened 2022-10-14
Improve performance of scrolling in song select

### LINKED ISSUE #3433: Fps tanks when using absolute scrolling in song select with a lot of maps
absolute scrolling = dragging middle mouse button
43542 maps

https://user-images.githubusercontent.com/22303902/159829794-66ea9adc-e63c-4341-8f9b-ac3bf8d63a43.mp4

Illuminati-CRAZ:
Might fix #3433

Prevent song select container sprites from updating their content when not close to the start or end of the scrolling animation.
This does cause the text and banners to repeat when scrolling a long distance, but generally doesn't look too obvious.


Swan:
The reason it lags when updating the content is because all of the text is cached. 

Instead of doing this - which is a pretty big hack which ruins the user experience, we could turn off caching for the text entirely or turn it off only while scrolling. The latter would be a bit more complex to do, but it will be the most performant.

Illuminati-CRAZ:
See #3737

PR #3672, opened 2022-10-16
add close button to automod panel

Kizuna-dev:
self-explanatory

### DIFF
--- a/Quaver.Shared/Screens/Edit/UI/AutoMods/EditorAutoModHeader.cs
+++ b/Quaver.Shared/Screens/Edit/UI/AutoMods/EditorAutoModHeader.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using Microsoft.Xna.Framework.Graphics;
 using Quaver.Shared.Assets;
 using Quaver.Shared.Helpers;
@@ -20,6 +20,8 @@ namespace Quaver.Shared.Screens.Edit.UI.AutoMods
 
         private SpriteTextPlus Text { get; set; }
 
+        private IconButton CloseButton { get; set; }
+
         public EditorAutoModHeader(EditorAutoModPanel panel)
         {
             Panel = panel;
@@ -29,6 +31,7 @@ namespace Quaver.Shared.Screens.Edit.UI.AutoMods
 
             CreateGearIcon();
             CreateText();
+            CreateCloseButton();
         }
 
         private void CreateGearIcon() => Gear = new Sprite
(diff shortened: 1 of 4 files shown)


Quaver.Shared/Screens/Edit/UI/AutoMods/EditorAutoModPanelContainer.cs
```diff
@@ -73,7 +73,8 @@ public EditorAutoModPanelContainer(EditScreen screen)
             Panel = new EditorAutoModPanel(Screen.WorkingMap, mapset)
             {
                 Parent = this,
-                Alignment = Alignment.MidRight
+                Alignment = Alignment.MidRight,
+                Container = this
```
Swan, line 77: This should be passed into the constructor instead. Without this line, the entire game will crash when clicking the button because of a null reference, so you may as well force it to be set.

Quaver.Shared/Screens/Edit/UI/AutoMods/EditorAutoModHeader.cs
```diff
@@ -49,5 +52,22 @@ private void CreateText()
...
+            var img = FontAwesome.Get(FontAwesomeIcon.fa_times);
+
+            CloseButton = new IconButton(img)
+            {
+                Parent = this,
+                Alignment = Alignment.MidRight,
+                X = -Height / 2,
+                Size = Gear.Size,
+                Depth = -1,
+                Tint = new Microsoft.Xna.Framework.Color(249, 100, 93, 255)
```
Swan, line 67: Any reason for including the entire namespace? Why not just `new Color()`?

Swan (changes requested):

Swan (approved):

Quaver.Shared/Screens/Edit/UI/AutoMods/EditorAutoModPanelContainer.cs
Kizuna-dev: Passing it as a constructor now, and force it to be set. I avoided doing that because ``AutoModTestScreenView`` also created a new panel and i didn't know what to pass, decided to pass ``Container as EditorAutoModPanelContainer`` and it work, hopefully that is the right way to do it

Swan: That technically is incorrect because the `Container` in the test is for the screen rather than the automod container (which is very specific to the editor). Doesn't matter too much for the test though. This is ok.

PR #3688, opened 2022-11-11
Create base keybind config system

IceDynamix:
## Keybind rules

- Left/Right modifiers keys are generalized into a single modifier variant
- Keybinds consist of one non-modifier key and any number of modifiers
- If the number of pressed modifiers does not exactly match, it will not match the keybind (Shift+Alt+S does not match Shift+S)
  - The "Free" modifier removes that condition and will match even if excess modifiers are pressed
  - If other modifiers are used along with the free modifier, the other modifiers must be pressed

## Configuration file

- Uses YAML format that (de)serializes the EditorInputConfig class
- File format is laid out in action->keybinds
- Uses custom serializing for the Keybind by parsing from/into its string representation using [KeybindYamlTypeConverter](https://github.com/Quaver/Quaver/pull/3688/files#diff-a0763781deb6d4468d823e2911a81583dc8c67c9694a90775d0f4f75f96453a6)
- Uses custom list style for the KeybindList deserialization using [KeybindListYamlFlowStyle](https://github.com/Quaver/Quaver/pull/3688/files#diff-d7c0b69eb6a9714403e11dbfa34044c3e826a8ec379aa640132a315142ee97e2)

## Keybind matching system

- Uses a generalized key state and gets all matched keybinds by comparing the current and previous state
  - Uses a reverse dictionary of keybind->actions to match the keybinds
  - A previous, more intuitive system that iterates over the action->keybinds dictionary and checks for unique keypresses for each action on each frame is inefficient, the FPS difference is 950 to 1050 on my setup

## Press/Hold/Release actions

- The time since the last keypress for an action and the time since the last execution of an action are recorded.
- Actions defined in `EditorInputManager.HoldRepeatActions`, are repeated in an interval after being held a certain amount of time. This is used in the future for seeking or zooming.
- Actions defined in `EditorInputManager.HoldAndReleaseActions` can be held and are executed on every frame the key is held down, and finally when the key is released. The states are mainly used for (long) note placement in the future.
- HandleAction() receives isPress and isRelease in order to check whether the currently executed action is a press, hold or release.

## Actions added in the subsequent updates

If actions are added in the future, they will not be added to the config automatically. An option will be provided to the user to fill missing actions either unbound or with the new default binds in the editor interface, using `EditorInputConfig.FillMissingKeys(bool fillWithDefaultBinds)`.

## Adding new keybind actions

1. Add new enum item in the `KeybindActions`
2. Add new default keybind in `EditorInputConfig.DefaultKeybinds`
3. Add new case in `EditorInputManager.HandleAction()`

### DIFF
--- a/Quaver.Shared/Screens/Edit/Input/EditorInputConfig.cs
+++ b/Quaver.Shared/Screens/Edit/Input/EditorInputConfig.cs
@@ -156,9 +156,10 @@ namespace Quaver.Shared.Screens.Edit.Input
                 .IgnoreUnmatchedProperties()
                 .Build();
 
-            EditorInputConfig config;
+            var config = ds.Deserialize<EditorInputConfig>(file);
 
-            config = ds.Deserialize<EditorInputConfig>(file);
+            if (config == null)
+                throw new Exception("Config file is empty");
 
             return config;
         }
--- a/Quaver.Shared/Screens/Edit/Input/EditorInputConfig.cs
+++ b/Quaver.Shared/Screens/Edit/Input/EditorInputConfig.cs
@@ -159,7 +159,10 @@ namespace Quaver.Shared.Screens.Edit.Input
             var config = ds.Deserialize<EditorInputConfig>(file);
 
             if (config == null)
-                throw new Exception("Config file is empty");
+            {
+                Logger.Debug("Config file was empty, creating new default", LogType.Runtime);
+                config = new EditorInputConfig();
+            }
 
             return config;
         }
--- a/Quaver.Shared/Screens/Edit/Input/EditorInputConfig.cs
+++ b/Quaver.Shared/Screens/Edit/Input/EditorInputConfig.cs
@@ -35,23 +35,19 @@ namespace Quaver.Shared.Screens.Edit.Input
             {
                 Logger.Debug("No editor key config found, using default", LogType.Runtime);
                 config.SaveToConfig();
+                return config;
             }
-            else
-            {
-                try
-                {
-                    using (var file = File.OpenText(ConfigPath))
-                    {
-                        config = Deserialize(file);
-                    }
 
-                    Logger.Debug("Loaded editor key config", LogType.Runtime);
-                    config.SaveToConfig(); // Reformat after loading
-                }
-                catch (Exception e)
-                {
-                    Logger.Error($"Could not load editor key config, using default: {e.Message}", LogType.Runtime);
-                }
+            try
+            {
+                using (var file = File.OpenText(ConfigPath))
+                    config = Deserialize(file);
+                Logger.Debug("Loaded editor key config", LogType.Runtime);
+                config.SaveToConfig(); // Reformat after loading
+            }
+            catch (Exception e)
+            {
+                Logger.Error($"Could not load editor key config, using default: {e.Message}", LogType.Runtime);
             }
 
             return config;
--- a/Quaver.Shared/Screens/Edit/Input/EditorInputConfig.cs
+++ b/Quaver.Shared/Screens/Edit/Input/EditorInputConfig.cs
@@ -1,0 +1,169 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using Quaver.Shared.Config;
+using Quaver.Shared.Graphics.Notifications;
+using Wobble.Logging;
+using Wobble.Platform;
+using YamlDotNet.Serialization;
+
+namespace Quaver.Shared.Screens.Edit.Input
+{
+    [Serializable]
+    public class EditorInputConfig
+    {
+        [YamlIgnore] public static string ConfigPath = ConfigManager.GameDirectory.Value + "/editor_keys.yaml";
+
+        public bool ReverseScrollSeekDirection { get; private set; }
+        public Dictionary<KeybindActions, KeybindList> Keybinds { get; private set; }
+        public Dictionary<string, KeybindList> PluginKeybinds { get; private set; }
+
+        public EditorInputConfig()
+        {
+            ReverseScrollSeekDirection = true;
+            Keybinds = DefaultKeybinds;
+            PluginKeybinds = new Dictionary<string, KeybindList>();
+        }
+
+        public static EditorInputConfig LoadFromConfig()
+        {
+            EditorInputConfig config;
+
+            if (!File.Exists(ConfigPath))
+            {
+                Logger.Debug("No editor key config found, using default", LogType.Runtime);
+                config = new EditorInputConfig();
+            }
+            else
+            {
+                using (var file = File.OpenText(EditorInputConfig.ConfigPath))
+                {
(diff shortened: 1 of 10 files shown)


AiAe (changes requested):
The config file is only created when entering Editor.
Unsure if this was the intended fix but having saved just "Ctrl+" its getting replaced by "Ctrl+None", probably better to just default it to None?

The keybinds needs to be more prone to bad syntax, for example if I do "[None] +" or "[None" editor wont load anymore.

Quaver.Shared/Screens/Edit/Input/EditorInputConfig.cs
```diff
@@ -0,0 +1,184 @@
...
+            if (Keybinds.ContainsKey(action))
+                return Keybinds[action].Remove(keybind);
+            return false;
+        }
+
+        public void SaveToConfig()
+        {
+            File.WriteAllText(ConfigPath, Serialize());
+            Logger.Debug("Saved editor key config to file", LogType.Runtime);
+        }
```
Swan, line 83: If this fails, the game will crash. In some cases you're handling it with a try/catch but in other cases you aren't. It either needs to be handled once here (ideally), or handled in each individual use case.

Quaver.Shared/Screens/Edit/Input/EditorInputConfig.cs
```diff
@@ -0,0 +1,184 @@
...
+                    Logger.Debug("Loaded editor key config", LogType.Runtime);
+                    config.SaveToConfig(); // Reformat after loading
+                }
+                catch (Exception e)
+                {
+                    Logger.Error($"Could not load editor key config, using default: {e.Message}", LogType.Runtime);
+                }
+            }
+
+            return config;
```
Swan, line 57:
```suggestion
            if (!File.Exists(ConfigPath))
            {
                Logger.Debug("No editor key config found, using default", LogType.Runtime);
                config.SaveToConfig();
                return config;
            }

            try
            {
                using (var file = File.OpenText(ConfigPath))
                    config = Deserialize(file);

                Logger.Debug("Loaded editor key config", LogType.Runtime);
                config.SaveToConfig(); // Reformat after loading
            }
            catch (Exception e)
            {
                Logger.Error($"Could not load editor key config, using default: {e.Message}", LogType.Runtime);
            }

            return config;
```

Just to reduce nesting a bit.

Quaver.Shared/Screens/Edit/Input/EditorInputConfig.cs
```diff
@@ -0,0 +1,184 @@
...
+
+            EditorInputConfig config;
+
+            try
+            {
+                config = ds.Deserialize<EditorInputConfig>(file);
+            }
+            catch (YamlException e)
+            {
+                throw new Exception($"Failed to parse configuration in line {e.Start.Line}");
```
Swan, line 160: Kinda defeats the purpose of the try catch if it's going to throw another exception? Maybe just return a fresh instance of the config and log the error.

Quaver.Shared/Screens/Edit/Input/EditorInputManager.cs
```diff
@@ -0,0 +1,150 @@
...
+            if (View.IsImGuiHovered) return;
+
+            HandleKeypresses();
+            HandleKeyReleasesAfterHoldAction();
+            HandlePluginKeypresses();
+
+            HandleMouseInputs();
+        }
+
+        private void HandleKeypresses()
```
Swan, line 62:
```suggestion
        private void HandleKeyPresses()
```

And all other instances of Presses not being capitalized.

Quaver.Shared/Screens/Edit/Input/EditorInputConfig.cs
```diff
@@ -0,0 +1,184 @@
...
+
+        public void OpenConfigFile()
+        {
+            try
+            {
+                Utils.NativeUtils.OpenNatively(ConfigPath);
+            }
+            catch (Exception e)
+            {
+                Logger.Debug(e.ToString(), LogType.Runtime);
```
Swan, line 93: Log it as an error.

Quaver.Shared/Screens/Edit/Input/KeybindActions.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System;
+
+namespace Quaver.Shared.Screens.Edit.Input
+{
+    [Serializable]
+    public enum KeybindActions
+    {
+        DebugAction
```
Swan, line 8: You can remove this action before this gets merged.

Swan (changes requested):

AiAe:
Writing the issues from our DMs:
Alert needs to be added for when it fails to parse/load the keybinds.
If the file is left empty the editor will still not open.

```
[9:00:54] - RUNTIME - ERROR: System.NullReferenceException: Object reference not set to an instance of an object.
   at Quaver.Shared.Screens.Edit.Input.EditorInputManager..ctor(EditScreen screen) in C:\Users\AiAe\Downloads\Quaver\Quaver.Shared\Screens\Edit\Input\EditorInputManager.cs:line 44
   at Quaver.Shared.Screens.Edit.EditScreen..ctor(Map map, IAudioTrack track, EditorVisualTestBackground visualTestBackground) in C:\Users\AiAe\Downloads\Quaver\Quaver.Shared\Screens\Edit\EditScreen.cs:line 289
   at Quaver.Shared.Screens.Selection.SelectionScreen.<>c__DisplayClass69_0.<ExitToEditor>b__0() in C:\Users\AiAe\Downloads\Quaver\Quaver.Shared\Screens\Selection\SelectionScreen.cs:line 714
   at Quaver.Shared.Screens.QuaverScreenManager.<>c__DisplayClass24_0.<ScheduleScreenChange>b__0() in C:\Users\AiAe\Downloads\Quaver\Quaver.Shared\Screens\QuaverScreenManager.cs:line 205
   at Quaver.Shared.Scheduling.ThreadScheduler.<>c__DisplayClass1_0.<Run>b__0() in C:\Users\AiAe\Downloads\Quaver\Quaver.Shared\Scheduling\ThreadScheduler.cs:line 38
```

Quaver.Shared/Screens/Edit/Input/EditorInputConfig.cs
```diff
@@ -156,9 +156,10 @@ private static EditorInputConfig Deserialize(StreamReader file)
                 .IgnoreUnmatchedProperties()
                 .Build();
 
-            EditorInputConfig config;
+            var config = ds.Deserialize<EditorInputConfig>(file);
 
-            config = ds.Deserialize<EditorInputConfig>(file);
+            if (config == null)
+                throw new Exception("Config file is empty");
```
Swan, line 162: This should recreate the config file for them instead of throwing an exception.

Swan (changes requested):

Issue #3694, opened 2022-11-19
offline


Abizzzi:
Please add offline mode , sometimes i have not internet connection
When I go to the game in offline mode, it comes, but when I want to go in a level, the screen is black

AiAe:
Click on your profile and click "logout"

Abizzzi:
> Click on your profile and click "logout"

 When I start Steam in offline mode, it doesn't work
It still shows a black screen

AiAe:
Go to Options > Misc and turn off "Automatically log into the server" 

Tested it and everything seems fine.

Abizzzi:
> Go to Options > Misc and turn off "Automatically log into the server"
> 
> Tested it and everything seems fine.

This way did not work.

What I did :

(1) no internet connection
(2) launch steam
(3) connection error ( start in offline mode )
(4) launch quaver
(5) play a map ( Already logged out )
(6) see error ( black screen )

Maybe this problem is due to my system , idk

OS : Windows 10

Swan:
@Abizzzi  Can't reproduce either. Will need your runtime.log and network.log files located in the Logs folder.

PR #3703, opened 2022-11-28
Run update ranked statuses only if online

AiAe:


Swan:
What's the reason for this PR? That endpoint doesn't require you to be online.

IceDynamix:
Dev client will report an error on updating ranked statuses every time you leave the editor otherwise

Swan:
> Dev client will report an error on updating ranked statuses every time you leave the editor otherwise

This is because there's no API server running locally? Not seeing any errors right now...

AiAe:
> > Dev client will report an error on updating ranked statuses every time you leave the editor otherwise
> 
> This is because there's no API server running locally? Not seeing any errors right now...

https://github.com/Quaver/Quaver/issues/3694
This happens when Steam is running in offline. 
It wont resolve the issue exactly but will be one less thing

PR #3749, opened 2023-02-16
Implement Editor Bookmarks

### LINKED ISSUE #659: Editor: Bookmarks
Bookmark support in the editor - This could be stored as metadata in the chart like in osu but I would like bookmarks to do a couple more things (QOL for modders and users alike)

- Bookmarks can be placed in the editor and there are two types

- The first type of bookmark is "note" which is used to add a remark or reminder for a specific point in the chart. It would be nice if you could place them by clicking next to the playfield and it places a bubble/message box next to where you clicked with a typeable box. The box then scrolls/moves to stay in line with the place it was placed at and is only visible when you're looking there in the chart.

- The next type of bookmark is "section divider" which can be used to place markers dividing sections of music apart e.g. most songs can be broken into sections with significant changes/transition sections at certain points. This can help with breaking down work you have to do, identifying similar sections, jumping to places quickly etc. Section divider bookmarks should also store text notes which could appear as a panel somewhere on the side of the editor when the cursor is within this divided section.

Other comments
- Bookmarks are drawn as vertical lines on the time navigation bar on the bottom (maybe dark blue for section dividers, light blue/aqua for notes?) and there is a way to quickly jump between both bookmarks (of both types)
- I don't know if you have some sort of protocol for this yet, but if/when chart modding comes around it would be nice if there was a way to click on a modding link and have the client scroll to that point in the editor, making a translucent "temporary" bookmark to show the comment at the point it is specifying
- As a follow up to this modding thing - It would be awesome to be able to export all "note" bookmarks into a block of formatted text to easily paste it as a mod on the site
- There is a shortcut to select the "currently divided section" (most recent divider point until earliest next divider point) for ease of quickly applying filters to divided sections/duplicating sections/various QOL

having notepad open on the side containing my ideas and stuff - this is what i normally do for modding and then format it later for posting

Swan:
Requires https://github.com/Quaver/Quaver.API/pull/142
Closes #659


Quaver.Shared/Screens/Edit/Dialogs/EditorBookmarkDialog.cs
```diff
@@ -0,0 +1,80 @@
...
+namespace Quaver.Shared.Screens.Edit.Dialogs
+{
+    public class EditorBookmarkDialog : YesNoDialog
+    {
+        private EditorActionManager ActionManager { get; }
+
+        private IAudioTrack Track { get; }
+        
+        /// <summary>
+        ///     The book mark that's currently being edited. If none is provided in the constructor,
```
AiAe, line 23: Should be "bookmark"

Illuminati-CRAZ (commented):
Would be nice to have plugins be able to interact with bookmarks

PR #3751, opened 2023-02-19
Fixes selector going below editor when preview is enabled

AiAe:


Quaver.Shared/Screens/Edit/EditScreenView.cs
```diff
@@ -393,6 +393,10 @@ private void CreateGameplayPreview()
             
             // Makes it so that the playfield bookmark tooltips appear above preview.
             Playfield.Parent = Container;
+            
+            // Makes it so the selector goes above editor after enabling preview.
+            if(Selector != null)
```
Swan, line 398: This null check shouldn't be required.

Swan (changes requested):

Quaver.Shared/Screens/Edit/EditScreenView.cs
AiAe: It does, without it it still throws error. Unsure about other solution to it.

Swan: What error does it throw? I've removed it and it works fine. The selector is created before the preview, so it should never be null.

AiAe: Cannot reproduce it anymore, but from what I remember it was "Object reference not set to an instance of an object"

AiAe: Well hopefully it wont be an issue, since when it was it prevented the editor to load at all and it was black screen.

Swan (approved):

PR #3763, opened 2023-03-01
Log response from server

AiAe:
Few people complained they have issues with uploading their mapsets. 
Would be good to have the response from the server in the logs for better debugging.


Quaver.Shared/Screens/Edit/Dialogs/EditorUploadingMapsetDialog.cs
```diff
@@ -59,6 +59,8 @@ public class EditorUploadingMapsetDialog : LoadingDialog
 
                 var path = MapManager.Selected.Value.Mapset.ExportToZip(false);
                 var response = OnlineManager.Client.UploadMapset(path);
+                
+                Logger.Important(response.ToString(), LogType.Network);
```
Swan, line 63: Did you test this? This doesn't look right.

Swan: Nvm, all good

Swan (approved):

PR #3783, opened 2023-04-14
Allow users with tournament mode privilege to spectate multis

Illuminati-CRAZ:
Requested by Aiae. Don't know if it works. Can't test because multiplayer. Also would be nice to add the privilege to the UserGroups enum but I can't edit it.


Swan:
@AiAe Is this something that needs to be added right now? If so then I'll merge it, but if not then I'd rather it be done the proper way which requires updating one of the client libraries.

AiAe:
> @AiAe Is this something that needs to be added right now? If so then I'll merge it, but if not then I'd rather it be done the proper way which requires updating one of the client libraries.

24th start the streams so i will need it by then.

Swan (approved):

Issue #3784, opened 2023-04-15
Local offset changes along playback rate in editor


PianoLuigi:
On Canary to reproduce:
- Have a long, empty intro before first notes in the map
- Enable gameplay preview

If you have local offset keys binded to + and - (which I think are the default), using ctrl +/- to change the song playback rate in the editor WHILE your cursor is located before the first notes on the timeline, the local offset will change in increments of 1, as the game assumes you are in gameplay.

PR #3795, opened 2023-05-01
Prevent offset change in editor

AiAe:
Fixes https://github.com/Quaver/Quaver/issues/3784

### DIFF
--- a/Quaver.Shared/Screens/Gameplay/GameplayScreen.cs
+++ b/Quaver.Shared/Screens/Gameplay/GameplayScreen.cs
@@ -1553,7 +1553,7 @@ namespace Quaver.Shared.Screens.Gameplay
             }
 
             // Only allow offset changes if the map hasn't started or if we're on a break
-            if (!IsSongSelectPreview && Ruleset.Screen.Timing.Time <= 5000 || Ruleset.Screen.EligibleToSkip)
+            if (!IsSongSelectPreview && !(this is EditScreen) && (Ruleset.Screen.Timing.Time <= 5000 || Ruleset.Screen.EligibleToSkip))
             {
                 var change = 5;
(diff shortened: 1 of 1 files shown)


Quaver.Shared/Screens/Gameplay/GameplayScreen.cs
```diff
@@ -1553,7 +1553,7 @@ public void HandleAutoplayTabInput(GameTime gameTime)
             }
 
             // Only allow offset changes if the map hasn't started or if we're on a break
-            if (!IsSongSelectPreview && Ruleset.Screen.Timing.Time <= 5000 || Ruleset.Screen.EligibleToSkip)
+            if (!IsSongSelectPreview && !(this is EditScreen) && (Ruleset.Screen.Timing.Time <= 5000 || Ruleset.Screen.EligibleToSkip))
```
Swan, line 1556: I don't think this line makes sense?

This will always equal true because `this` will always be equal to`GameplayScreen`

Swan (changes requested):

Quaver.Shared/Screens/Gameplay/GameplayScreen.cs
Swan: And this case should already be covered by the `!IsSongSelectPreview`?

AiAe: From my testing `!IsSongSelectPreview` did not cover it. 
And my fix did fix it from my testing, but maybe I missed something? 
Will recheck later.

Swan: If it doesn't, then it should. Because ideally it would disable offset changes anywhere a map preview is active.

Issue #3808, opened 2023-06-13
Editor: Specific BPMS prevent scrolling at random places


naether:
**Describe the bug**
At some BPMs, at any beat snap that is a multiple of 2, the editor stops scrolling up at seemingly random places
Using any other beat snap, or scrolling while the song is playing works fine, and does scrolling down

**To Reproduce**
1. Create a blank chart with any song
2. Add a 242 BPM timing point (or any other broken bpm)
3. Scroll up until you suddenly can't!!!! what the hll!!!!!!

**Expected behavior**
can scroll up

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser: Chrome
 - Version: Latest (v1.1)

**Additional context**
 234.38 also is broken, i have no idea what causes this

PR #3811, opened 2023-06-18
Fix editor not scrolling due to rounding errors

WarpABoi:
fixes bug described in #3808


Quaver.Shared/Audio/AudioEngine.cs
```diff
@@ -173,7 +173,7 @@ public static double GetNearestSnapTimeFromTime(Qua map, Direction direction, in
 
             var nearestTick = Math.Round((pointToSnap - point.StartTime) / snapTimePerBeat) * snapTimePerBeat + point.StartTime;
 
-            if ((int) Math.Abs(nearestTick - time) <= (int) snapTimePerBeat)
+            if ((int)Math.Abs(nearestTick - time) <= (int)snapTimePerBeat || Math.Abs(nearestTick - time) - snapTimePerBeat <= 0.1)
```
YaLTeR, line 176: So I'm not exactly quite sure what this condition is doing (why is it rounding to ints?), but either way this seems like an odd way to correct it. Can't you just replace the condition with something along `if (Math.Abs(nearestTick - time) <= snapTimePerBeat + 0.1)`? Also, is that 0.1 constant good, or maybe it should be smaller?

WarpABoi: it might be better to change that value depending on the bpm, i can try some things later

Swan (approved):

PR #3812, opened 2023-06-21
Add and implement skinning Judgement Burst Pos X

Syafiqjos:
Add JudgementBurstPosX to be modified on skin.ini.
JudgementBurstPosX  is the X Offset of the judgement hit burst.
Only take an effect when DisplayJudgementsInEachColumn is false.

### DIFF
--- a/Quaver.Shared/Screens/Gameplay/Rulesets/Keys/Playfield/GameplayPlayfieldKeysStage.cs
+++ b/Quaver.Shared/Screens/Gameplay/Rulesets/Keys/Playfield/GameplayPlayfieldKeysStage.cs
@@ -565,7 +565,7 @@ namespace Quaver.Shared.Screens.Gameplay.Rulesets.Keys.Playfield
                 {
                     Parent = Playfield.ForegroundContainer,
                     Alignment = Alignment.MidCenter,
-                    X = skin.DisplayJudgementsInEachColumn ? Receptors[lane].X - playfieldOffset : 0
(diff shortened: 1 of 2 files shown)


Quaver.Shared/Screens/Gameplay/Rulesets/Keys/Playfield/GameplayPlayfieldKeysStage.cs
```diff
@@ -565,7 +565,7 @@ private void CreateJudgementHitBurst()
                 {
                     Parent = Playfield.ForegroundContainer,
                     Alignment = Alignment.MidCenter,
-                    X = skin.DisplayJudgementsInEachColumn ? Receptors[lane].X - playfieldOffset : 0
+                    X = skin.DisplayJudgementsInEachColumn ? Receptors[lane].X - playfieldOffset : (skin.JudgementBurstPosX)
```
YaLTeR, line 568: Nitpick: I'd rather enclose in () the expression to the left of : as it is multi-part, rather than this.

YaLTeR (approved):

Quaver.Shared/Screens/Gameplay/Rulesets/Keys/Playfield/GameplayPlayfieldKeysStage.cs
Syafiqjos: You're right, I'll fix it

YaLTeR (approved):

Issue #3745, opened 2023-01-28
Reverse/Inverse Scoring


Storn42:
**Is your feature request related to a problem? Please describe.**
Sometimes its just nice to have different ways to track score during a song, and it can be hard to understand the exact impact of getting a bad judgement while playing as it only considers the notes hit so far. 

**Describe the solution you'd like**
Instead of Score being based on the notes you've already hit and starting at 0, An inverse scoring system starts with a perfect score and subtracts from it every time you make a non perfect (non-marv) judgement. This could be used on score, accuracy, and rating, as well as the end score graphs. This is a visual change and doesn't impact the end result of a play, only the visual feedback while playing.

**Describe alternatives you've considered**
There really isn't any.

**Additional context**
Scoring like this is very common in many rhythm games. Chrono Circle, Wacca, USC, FFR (kind of) all have Inverse scoring systems.

PR #3861, opened 2023-08-12
add inverse scoring

WarpABoi:
adds #3745


Swan:
Not sure if this should be added to the game. 

Score itself is a pretty useless metric that we've kept in the game but haven't done anything with.  I don't see the benefit of building on top of something that already doesn't provide much value.

PR #3864, opened 2023-08-16
add option to disable recommend difficulty dialog

WarpABoi:
this dialog has started annoying me

i dont know if i chose the best names for the options, feel free to suggest new ones


Swan:
Please open up issues before pull requests, so no time is wasted adding features that won't make it into the game.

PR #3866, opened 2023-08-16
Ghost lobby fix

### LINKED ISSUE #3865: Player can leave from multiplayer game but still be there
### Describe the bug

If your host and pick an map, then press the Home button from navbar it will get you out of the lobby, but it will still count you in it + chat. 

### To Reproduce

1. Create game
2. Go to selecting map
3. Click on Home
4. Go back to multiplayer games you will see 1/16 players
5. Clicking spectating you will see yourself in the lobby

### Expected behavior

_No response_

### Log Files

n/a

### Screenshots

_No response_

WarpABoi:
most likely fixes the issue, tho im unable to test multiplayer stuff.

fixes #3865

### DIFF
--- a/Quaver.Shared/Graphics/Menu/Border/Components/Buttons/IconTextButtonHome.cs
+++ b/Quaver.Shared/Graphics/Menu/Border/Components/Buttons/IconTextButtonHome.cs
@@ -2,6 +2,7 @@
 using Microsoft.Xna.Framework;
 using Microsoft.Xna.Framework.Graphics;
 using Quaver.Shared.Assets;
+using Quaver.Shared.Online;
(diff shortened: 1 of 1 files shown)


Swan (changes requested):
This is an odd way of doing this.

To solve the issue listed, simply prevent the user from going to that screen if they're in a multiplayer game rather than leaving the game entirely.

WarpABoi:
when in the multiplayer screen it allows you to exit with the home button in the header.

wouldnt it be inconsistent to restrict it in the song select?

Swan:
The elegant solution would be to leave the game when the user clicks on the home button.

Swan (approved):

PR #3895, opened 2023-09-03
dont limit characters to 11 in layer change color dialog

WarpABoi:
its a crude way to limit to rgb


Swan:
Could you explain the following:

- Why it's a "crude way" to limit rgb? 
- Why the PR removes the limit rather than fixes the issue (if one exists) at hand?

WarpABoi:
if there is spaces in the input like "255, 255, 255" it would be cut to "255, 255, 2"

when it successfully parses the input it replaces it without spaces

Swan (approved):

PR #3916, opened 2023-10-06
✨ Prepare Localization

GreenScreen410:


Quaver.Shared/Config/ConfigManager.cs
```diff
@@ -793,6 +793,11 @@ public static class ConfigManager
         /// </summary>
         internal static Bindable<GameMode> PrioritizedGameMode { get; private set; }
 
+        /// <summary>
+        ///     Change the language of the game.
+        /// </summary>
+        internal static Bindable<Language> ChangeLanguage { get; private set; }
```
Warp9000, line 799: Shouldnt this just be named Language

AiAe:
Marking this PR as draft, since its not ready to be merged.
Also to comment on this idea, would suggest if you wait for this to be discussed internally before more effort into this is put.

PR #3951, opened 2024-01-24
Add support for per-lane hitlights.

SSC1969:
Introduce the ability to skin hit and holdlights for individual lanes using the file naming convention "hitlighting-{lane}@{rows}x{columns}.png". If a file with the previous method for lighting is present, it will be loaded for any lanes that do not have a dedicated lighting spritesheet, allowing for full support of currently existing skins.

### DIFF
--- a/Quaver.Shared/Skinning/SkinKeys.cs
+++ b/Quaver.Shared/Skinning/SkinKeys.cs
@@ -355,12 +355,12 @@ namespace Quaver.Shared.Skinning
         /// <summary>
         ///
         /// </summary>
-        internal List<Texture2D> HitLighting { get; private set; } = new List<Texture2D>();
+        internal List<List<Texture2D>> NoteHitLighting { get; } = new List<List<Texture2D>>();
 
         /// <summary>
         ///
         /// </summary>
-        internal List<Texture2D> HoldLighting { get; private set; } = new List<Texture2D>();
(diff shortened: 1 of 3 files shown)


Quaver.Shared/Skinning/SkinKeys.cs
```diff
@@ -674,6 +672,34 @@ private void LoadHitObjects(IList<List<Texture2D>> hitObjects, string element, i
...
+        {
+            var mode = ModeHelper.ToShortHand(Mode).ToLower();
+
+            if (!Directory.Exists($"{Store.Dir}/{mode}"))
+                return;
+            
+            var folder = $"{Store.Dir}/{mode}/Lighting";
+
+            // Check for lane-specific lighting; otherwise, load general lighting for the lane
+            if (Directory.EnumerateFiles(folder).Any(f => f.Contains($"{type}-{lane}")))
```
Warp9000, line 692: Naming a file to something like hitlighting-2abc.png would cause this to try to loadhitlighting-2.png even though it does not exist.

SSC1969: Not sure how I missed that, updated the code to use Regex.IsMatch to guarantee the correct file is searched for.

AiAe:
Closing due to inactivity

PR #3964, opened 2024-02-21
Beat snap dialog enhancements

Emik03:
This PR allows users to quickly add multiple beat snaps by specifying a range with the syntax `x-y`, while still allowing the original syntax `x`. (where `x` and `y` are beat snap values)

### Why
If you have a song that uses many different beat snaps, you have to normally add them one-by-one, which can be time-consuming, especially when you don't know the beat snaps upfront and have to experiment to see which one aligns the best.

### Edgecases (and how this PR addresses them)
- Should the upper range be inclusive or exclusive?
    - While it is more common in programming languages for exclusive to be the default, or have separate syntax for both, doing so would add unnecessary complexity. A majority of the time in human language, people usually refer to inclusive bounds, such as asking "How was it, from a scale of 1 to 10?".
- How should a trailing `-` be handled? Example: `2-`.
    - Ignore it, treat it as `2`.
- How should a backwards range be handled? Example: `10-9`
    - Treat it as `9-10`, since some people refer to backwards ranges this way.
- How should the same number for upper and lower bound be handled?
    - While weird, I don't see why we should outright disallow this.
- Which beat snap should be selected if multiple are specified?
    - Always the lowest one. This choice is 100% arbitrary.
- What about when all beat snaps are selected? It takes up more vertical space than the window size.
    - ~~Currently this is unhandled, just as it was before this update. It isn't a deal breaker since CTRL+Scroll Wheel exists, but I would imagine that we may want to consider splitting into 2 columns when the amount of options grows past 24.~~ This is now handled by making the beat snap dialog scrollable.

### Demo
pr


Swan (dismissed):
> What about when all beat snaps are selected?

You can make the dropdown scrollable to fix this. See [OptionsItemScreenResolution](https://github.com/Quaver/Quaver/blob/ui-redesign/Quaver.Shared/Screens/Options/Items/Custom/OptionsItemScreenResolution.cs#L33-L39) as an example.

Warp9000:
You can now make it scrollable with `Dropdown.MaxHeight`. m´Make sure you pull the latest commits from ui-redesign.

Warp9000:
https://github.com/Quaver/Quaver/assets/59202433/1cf7231a-d60b-49a6-80ef-4c09db7ddbdf

The dropdown still scrolls when not hovering it.

AiAe (approved):

PR #3968, opened 2024-03-01
Add an Option for Automatically Enabling NF Mod upon Failure

QiuFeng54321:
If one just wishes to enjoy the music and isn't sure if they would pass the map, but would want to:
+ Still be able to submit their scores if they don't die
+ Still be able to play even if they die without `NF`

Then this could be very helpful.

+ [x] Add relevant config and option item
+ [x] Add handling code for enabling NF
+ [x] Disable NF upon leaving when it had been enabled during gameplay
+ + [x] Works when `Skip Results Screen` is enabled
+ + [x] Works when retrying
+ + [x] Does not affect Calibration view
+ + [x] Does not clear the NF mod if it is intentionally activated by the player
+ [x] NoMiss mod would take priority, such that it would restart the game upon missing when AutoNoFail is set
+ [x] Would not affect multiplayer (Well I couldn't check this but there's logic blocking it)
+ [x] Find a suitable place for the option item (please help me devs, I can't find an appropriate category so I just put it below 'Tap to Pause' for now)

### DIFF
--- a/Quaver.Shared/Screens/Gameplay/GameplayScreen.cs
+++ b/Quaver.Shared/Screens/Gameplay/GameplayScreen.cs
@@ -983,7 +983,7 @@ namespace Quaver.Shared.Screens.Gameplay
             {
                 // Add no fail mod upon dying when AutoNoFail is on.
                 // Add the no fail mod to their score.
-                NotificationManager.Show(NotificationLevel.Warning, "WARNING! Your score will not be submitted due to dying " +
+                NotificationManager.Show(NotificationLevel.Warning, "WARNING! Your score will not be submitted due to failing " +
                                                                     "during gameplay!", null, true);
                 IsNoFailAddedInGameplay = true;
             }
--- a/Quaver.Shared/Screens/Gameplay/GameplayScreen.cs
+++ b/Quaver.Shared/Screens/Gameplay/GameplayScreen.cs
@@ -195,6 +195,10 @@ namespace Quaver.Shared.Screens.Gameplay
         ///     If the user quit the game themselves.
         /// </summary>
         public bool HasQuit { get; set; }
+        /// <summary>
+        ///     If the no-fail mod is added during gameplay, when the player dies
+        /// </summary>
+        public bool IsNoFailAddedInGameplay { get; set; }
 
         /// <summary>
         ///     Flag that dictates if the user is currently restarting the play.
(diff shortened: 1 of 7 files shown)


QiuFeng54321:
Starting in [0a7b786](https://github.com/Quaver/Quaver/pull/3968/commits/0a7b7868edffb1d0bbde19366d2590bf8b2eaa08), a different logic is used, as described in the commit message. The reason for this change is that `Failed` is being used 20 times, and it being `true` for even a millisecond could potentially cause other code using this as a condition to trigger. Now, `Failed` is `false` when `AutoNoFail` is set and the player fails. A separate logic in `HandleFailure()` is used for the enabling of NF instead. This logic is the original logic of `Failed` but with some modifications. I need advises in potentially merging the two logic in some way of extracting some common part out of it, so that further modifications related to the failing condition would not need to take care of both logic at the same time.

Quaver.Shared/Screens/Gameplay/GameplayScreen.cs
```diff
@@ -527,6 +536,8 @@ public override void Update(GameTime gameTime)
             QuaverScreenChangeType type = QuaverScreenChangeType.CompleteChange)
         {
             Utils.NativeUtils.EnableWindowsKey();
+            if (IsNoFailAddedInGameplay)
+                ModManager.RemoveMod(ModIdentifier.NoFail);
```
Warp9000, line 540: Removing NoFail here will make result screen try to submit the score online, you should remove it after it has checked for unranked mods.

Quaver.Shared/QuaverGame.cs
```diff
@@ -947,7 +947,6 @@ public static void SetAudioDevice(bool reloadResources = false)
             {
                 if (ConfigManager.AudioOutputDevice.Value != Bass.GetDeviceInfo(i).Name)
                     continue;
-
```
Warp9000, line 950: Whats with the random whitespace removal?

Warp9000 (changes requested):

Quaver.Shared/QuaverGame.cs
QiuFeng54321: Fixed

Quaver.Shared/Screens/Gameplay/GameplayScreen.cs
QiuFeng54321: Fixed. Now removes NF only when retrying, quitting from pause menu and after calling SubmitOnlineScore()

AiAe:
I was discussing with @Warp9000 wont it be better if we add new option for this, instead applying the mod mid game?
As in, players can toggle if after they fail they want to continue playing, and then we can just apply the NF mod at the end of the play? 
Wont this solution be better and safer than what your doing?

QiuFeng54321:
> I was discussing with @Warp9000 wont it be better if we add new option for this, instead applying the mod mid game? As in, players can toggle if after they fail they want to continue playing, and then we can just apply the NF mod at the end of the play? Wont this solution be better and safer than what your doing?

I changed it so that it applies the NF mod when the Results Screen loads and before the score submits. 

Also I made an option subcategory `Gameplay->Others` and put the toggle there. I don't really think it could be fit in anywhere.

QiuFeng54321:
Closing due to too much commits containing abandoned features. Reopening a new one.

PR #3976, opened 2024-03-09
Make Displaying Pausing Warning Toggleable

QiuFeng54321:
Some rages when seeing this notification, so I made it toggleable.

### DIFF
--- a/Quaver.Shared/Config/ConfigManager.cs
+++ b/Quaver.Shared/Config/ConfigManager.cs
@@ -609,6 +609,10 @@ namespace Quaver.Shared.Config
         /// <summary>
         /// </summary>
         internal static Bindable<bool> DisplayNotificationsInGameplay { get; private set; }
+        
+        /// <summary>
+        /// </summary>
+        internal static Bindable<bool> DisplayPauseWarning { get; private set; }
 
         /// <summary>
         /// </summary>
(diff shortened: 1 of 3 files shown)


YaLTeR:
Do we really need to take up space in the options menu with things like this? It's hard enough to find what you're looking for as is.

QiuFeng54321:
> Do we really need to take up space in the options menu with things like this? It's hard enough to find what you're looking for as is.

Maybe we could put this option in Misc->Notification?

It does seem inappropriate to use an option item for just this, but I don't have any idea how else we could toggle it. Do you have any advise?

YaLTeR:
Idk, could leave as ini-only, since the setting sounds quite specific.

QiuFeng54321:
> Idk, could leave as ini-only, since the setting sounds quite specific.

OK, I removed the option item.

Illuminati-CRAZ:
lgtm, I might add a similar setting for the quick exit setting to the pr

Illuminati-CRAZ (approved):

AiAe:
I think we should avoid having "hidden" config settings, everything should be always available in-game.

QiuFeng54321:
> I think we should avoid having "hidden" config settings, everything should be always available in-game.

I added an option item for it in `Options->Misc->Notifications`.

Warp9000 (approved):

PR #3978, opened 2024-03-12
Change default fps limiter to custom and set it to 1000

AiAe:
We should not have the default FPS set to unlimited for new players.


Illuminati-CRAZ:
I feel 1000 fps default is too much. Like I can reach 2000 fps in gameplay when unlimited, but I have mine limited to 660 (4x my monitor refresh rate) to minimize stuttering.

Thoughts on 4x monitor refresh rate, and if that's greater than 500 fps, then the greatest multiple of monitor refresh rate less than 500 fps?

60 hz monitor (potentially underpowered system) => 240 fps limit
120 hz monitor => 480 fps limit
144 hz monitor => 432 fps limit
165 hz monitor => 495 fps limit
240 hz monitor => 480 fps limit
300 hz monitor => 300 fps limit (well, I guess this one's a bit low)

PR #3982, opened 2024-03-20
Migrate to .NET 8

Emik03:
This pull request migrates the project to .NET 8, making the same change of swapping `BinaryFormatter` with `XmlSerializer` in a copy of the function. (see https://github.com/Quaver/Quaver.API/pull/153)

Compiled successfully and ran without issues on Linux. Performance should theoretically be improved, but I haven't noticed so far.

### DIFF
--- a/Quaver/Quaver.csproj
+++ b/Quaver/Quaver.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <OutputType>WinExe</OutputType>
-    <TargetFramework>net6.0</TargetFramework>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net8.0</TargetFramework>
     <ApplicationIcon>Icon.ico</ApplicationIcon>
     <Description>
     </Description>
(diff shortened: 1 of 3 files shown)


AiAe:
The github workflow also needs to be updated to test for it. https://github.com/Quaver/Quaver/blob/ui-redesign/.github/workflows/dotnet.yml#L22

Illuminati-CRAZ:
> Performance should theoretically be improved, but I haven't noticed so far.

My assumption is that the .NET 8 performance improvements affect CPU, but it's the GPU that bottlenecks performance for Quaver.

Quaver/Quaver.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <OutputType>WinExe</OutputType>
-    <TargetFramework>net6.0</TargetFramework>
+    <OutputType>Exe</OutputType>
```
Illuminati-CRAZ, line 3: Not sure why the output type is being changed. By changing it from WinExe to Exe a console window is showing up on game startup.

Illuminati-CRAZ (changes requested):
Could you have Quaver.Shared target .net8 and Quaver.API target .net6 (for Quaver.API this should be as simple as updating the submodule to use the master branch)

AiAe:
Will close this PR for now since it also has breaking change

PR #4007, opened 2024-04-04
Switch Difficulty from within the Editor

### LINKED ISSUE #4004: Enable Switching Difficulties from the Editor Menu

every time you want to edit a different difficulty in the same mapset you have to exit the editing screen to the song select menu, select the mapset, right click the difficulty and edit it from there. this problem can be extra tedious because its possible to instead right click the mapset itself, which makes you edit the first difficulty. this can be really annoying when you consider loading times.

just add an option to swap to anther difficulty from the File menu, similar to the Reference Difficulty option.

I'm not creative enough

QiuFeng54321:
Resolves #4004


Quaver.Shared/Screens/Edit/Dialogs/UnsavedChangesSwitchMapDialog.cs
Warp9000, line 1: You should change the no image to decline and add a separate cancel button for canceling the map switch.

Warp9000 (changes requested):

Warp9000 (approved):

PR #4008, opened 2024-04-04
Fix order of parameters of Vector4 in CLR <-> Lua script conversion

### LINKED ISSUE #3998: `imgui.InputFloat4` does not have separate inputs
# Bug: When making a plugin, changing one input in `InputFloat4` changes all inputs simultaneously.

## MWE Example 1 (Faulty Behavior)
```lua
function draw()
	vars = state.GetValue("vars") or {0, 0, 0, 0}

	imgui.Begin("window")

	_, vars = imgui.InputFloat4("inputs", vars)

	state.SetValue("vars", vars)

	imgui.End()
end
```

### Simulation:
1. Inputs are `0.000` | `0.000` | `0.000` | `0.000`
2. Change the first input to `1.000`
3. All other inputs are now `1.000`
4. Resulting input list is `1.000` | `1.000` | `1.000` | `1.000` 

## MWE Example 2 (Expected Behavior)
```lua
function draw()
	vars = state.GetValue("vars") or {0, 0, 0}

	imgui.Begin("window")

	_, vars = imgui.InputFloat3("inputs", vars)

	state.SetValue("vars", vars)

	imgui.End()
end
```

### Simulation:
1. Inputs are `0.000` | `0.000` | `0.000`
2. Change the first input to `1.000`
4. Resulting input list is `1.000` | `0.000` | `0.000`

## MWE Example 3 (Expected Behavior)
```lua
function draw()
	vars = state.GetValue("vars") or {0, 0}

	imgui.Begin("window")

	_, vars = imgui.InputFloat2("inputs", vars)

	state.SetValue("vars", vars)

	imgui.End()
end
```

### Simulation:
1. Inputs are `0.000` | `0.000`
2. Change the first input to `1.000`
4. Resulting input list is `1.000` | `0.000`

### I expect this is due to an ID conflict that's made between the four inputs when `imgui.InputFloat4` is called.

QiuFeng54321:
Resolves #3998 
The order of Vector4 constructor is `x, y, z, w` and in lua conversion it constructs it as `w, x, y, z` instead. This makes the parameters keep cycling around and causes the issue. I fixed the ordering of the parameters.

### DIFF
--- a/Quaver.Shared/Scripting/LuaImGui.cs
+++ b/Quaver.Shared/Scripting/LuaImGui.cs
@@ -256,21 +256,21 @@ namespace Quaver.Shared.Scripting
             Script.GlobalOptions.CustomConverters.SetScriptToClrCustomConversion(DataType.Table, typeof(Vector4),
                 dynVal => {
                     var table = dynVal.Table;
-                    var w = (float)((double)table[1]);
-                    var x = (float)((double)table[2]);
-                    var y = (float)((double)table[3]);
-                    var z = (float)((double)table[4]);
-                    return new Vector4(w, x, y, z);
+                    var x = (float)((double)table[1]);
+                    var y = (float)((double)table[2]);
+                    var z = (float)((double)table[3]);
+                    var w = (float)((double)table[4]);
+                    return new Vector4(x, y, z, w);
                 }
             );
 
             Script.GlobalOptions.CustomConverters.SetClrToScriptCustomConversion<Vector4>(
                 (script, vector) => {
-                    var w = DynValue.NewNumber(vector.W);
                     var x = DynValue.NewNumber(vector.X);
                     var y = DynValue.NewNumber(vector.Y);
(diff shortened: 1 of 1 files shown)


Warp9000:
does this break any existing plugins?

QiuFeng54321:
> does this break any existing plugins?

If we want compatibility we could instead keep vector4 in lua w, x, y, z and only fix the cycling bug?

QiuFeng54321:
image

I checked the plugins I have. AmoguSV, Displacer, IceSV (Cubic Bezier was never set by user in float4 and is handled inside Lua using string), ParallelUniverse, KeepStill work fine with this change.

ESV-Sweetplum:
Coming from an SV nerd, no SV plugin I've tried uses `InputInt4` or `InputFloat4`. The purpose of me using `InputFloat4` is to avoid dealing with cubic bezier, and instead have 4 float-type coefficients that represent a polynomial.

Warp9000 (approved):

PR #4036, opened 2024-04-11
Fix hit position line again

QiuFeng54321:


QiuFeng54321:
I messed up rebase so reopening at #4042

PR #4038, opened 2024-04-11
Improvements to MapsetImporter

### LINKED ISSUE #4037: Merging of StepMania files when bulk importing with same .sm file names
### Describe the bug

https://github.com/Qiufeng54321/Quaver/blob/c422d078f7bf704c02df1697b2b9ff8b677b3c06/Quaver.Shared/Database/Maps/MapsetImporter.cs#L307

Problem happens here, when multiple SM folders are dragged to quaver window to be scheduled to import, when it actually imports, the created folder name for an SM mapset with `{filename}.sm` would be `{filename} - {timeMs % 1000}`. The folder name would be the same because:
+ they're parallelly executed, the timeMs % 1000 is the same
+ they are all called `file.sm`

So they ended up being merged under one folder

### To Reproduce

Queue import SM files whose file names are the same (e.g. all named `file.sm`)

### Expected behavior

_No response_

### Log Files

Not suitable for this issue

### Screenshots

_No response_

QiuFeng54321:
Resolves #4037 and #4052

`.mc` file import actually imports every other `.mc` files present under the same directory, so I made it to import only one of them. This improves the importing speed of e.g. dan mapsets.

If there are multiple StepMania mapsets being imported and they all have a file named `file.sm`, when bulk importing they are placed in the same directory (`file - 123` where 123 is replaced by the current millisecond). This PR changes this logic so it instead uses the directory name of the `.sm` files.

### DIFF
--- a/Quaver.Shared/Database/Maps/MapsetImporter.cs
+++ b/Quaver.Shared/Database/Maps/MapsetImporter.cs
@@ -139,6 +139,8 @@ namespace Quaver.Shared.Database.Maps
         /// <param name="path"></param>
         private static void AddMapImportToQueue(string path)
         {
+            // Only one .mc file under the same directory should be imported
+            // since .mc import 
             if (Path.GetExtension(path) == ".mc")
             {
                 foreach (var scheduledPath in Queue)
@@ -146,6 +148,7 @@ namespace Quaver.Shared.Database.Maps
                     if (Path.GetDirectoryName(scheduledPath) == Path.GetDirectoryName(path)) return;
                 }
             }
+
             NotificationManager.Show(NotificationLevel.Info, $"Scheduled {Path.GetFileName(path)} to be imported!");
             Queue.Add(path);
             PostMapQueue();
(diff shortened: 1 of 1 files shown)


Quaver.Shared/Database/Maps/MapsetImporter.cs
```diff
@@ -134,6 +133,24 @@ private static bool AcceptedMapType(string path)
...
+        private static void AddMapImportToQueue(string path)
+        {
+            if (Path.GetExtension(path) == ".mc")
+            {
+                foreach (var scheduledPath in Queue)
+                {
+                    if (Path.GetDirectoryName(scheduledPath) == Path.GetDirectoryName(path)) return;
+                }
+            }
+            NotificationManager.Show(NotificationLevel.Info, $"Scheduled {Path.GetFileName(path)} to be imported!");
```
AiAe, line 149: Add new line

QiuFeng54321:
Example: https://1drv.ms/u/c/45ffa70ba2f6f548/ETr_Zgwu3z9It5KcyoC5HEIBcj3auJAAnSNUpVd96A3JwA?e=69Xo2g

AiAe (approved):

PR #4054, opened 2024-04-13
Add option to scale ImGui in editor

QiuFeng54321:
Requires https://github.com/Quaver/Wobble/pull/139

The only issue I see with this is this:

image

There is a little gap between the menu bar and its options.


Quaver.Shared/Config/ConfigManager.cs
```diff
@@ -1064,6 +1069,7 @@ private static void ReadConfigFile()
             TapToRestart = ReadValue(@"TapToRestart", false, data);
             DisplayFailedLocalScores = ReadValue(@"DisplayFailedLocalScores", true, data);
             EditorScrollSpeedKeys = ReadInt(@"EditorScrollSpeedKeys", 16, 5, 100, data);
+            EditorImGuiScalePercentage = ReadInt(@"EditorImGuiScalePercentage", 100, 25, 400, data);
```
AiAe, line 1072: Change max to 300

AiAe (approved):

PR #4075, opened 2024-04-22
Fix audio seeking issues in map preview

QiuFeng54321:
* The audio track instance is not properly updated when the track ends. This causes the 2fps issue
*  * I think we should do some tests as to why this would cause the framedrop dramatically
* If dragging to the very end of the audio and not let go, the audio will keep being reloaded, seeked to the end and reloaded and so on and so forth. This also dramatically drops the fps. Now, we refuse to seek to within the last second of the audio in map preview seek bar.

### DIFF
--- a/Quaver.Shared/Screens/Gameplay/Rulesets/Keys/HitObjects/HitObjectManagerKeys.cs
+++ b/Quaver.Shared/Screens/Gameplay/Rulesets/Keys/HitObjects/HitObjectManagerKeys.cs
@@ -75,6 +75,11 @@ namespace Quaver.Shared.Screens.Gameplay.Rulesets.Keys.HitObjects
         private Qua Map;
 
         /// <summary>
+        ///     Cached length of the map
+        /// </summary>
+        private int Length { get; set; }
+
+        /// <summary>
         ///     Number of lanes
         /// </summary>
         private int KeyCount { get; }
@@ -239,7 +244,7 @@ namespace Quaver.Shared.Screens.Gameplay.Rulesets.Keys.HitObjects
                     return false;
 
                 // Wait for dead LNs to finish scrolling
-                return CurrentVisualAudioOffset > Map.Length;
+                return CurrentVisualAudioOffset > Length;
             }
         }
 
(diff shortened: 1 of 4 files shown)


A6sT:
There's still an issue on longer maps if the end of the map is empty (audio file keeps on playing but there's no more note to hit) that will cause the game to have huge fps drops. 
This can easily be replicated by taking any maps that is 4+ minutes long. There is a chance that this is not due to the length of the map but more because of the number of note on the map

Here's an example
https://github.com/Quaver/Quaver/assets/58571281/b1923856-aeb4-4e8e-909f-a79d7a7b2b72

QiuFeng54321:
> There's still an issue on longer maps if the end of the map is empty (audio file keeps on playing but there's no more note to hit) that will cause the game to have huge fps drops. This can easily be replicated by taking any maps that is 4+ minutes long. There is a chance that this is not due to the length of the map but more because of the number of note on the map
> 
> Here's an example https://github.com/Quaver/Quaver/assets/58571281/b1923856-aeb4-4e8e-909f-a79d7a7b2b72

Hi, can you try setting the FPS to Unlimited and see if you can still see this effect? I found that using Custom FPS Limiter will trigger this fps drop

A6sT:
This is indeed not replicable with unlimited fps. I have also checked with every frame limiter, and it seems that Custom is the only one that cause the game to struggle. Also, it looks like the higher the custom value is, the bigger the lag is

Quaver.Shared/Graphics/Graphs/DifficultySeekBar.cs
```diff
@@ -120,7 +120,8 @@ public override void Update(GameTime gameTime)
                 var percentage = (MouseManager.CurrentState.Y - AbsolutePosition.Y) / AbsoluteSize.Y;
                 var targetPos = (1 - percentage) * Track.Length;
 
-                SeekToPos(targetPos);
+                if (Track.Length - targetPos > 1000)
```
AiAe, line 123: Can this now also be removed? It kinda prevents you skipping to the end of the track in seekbar. Which is minor thing but I tested removing it, it does not change anything.

QiuFeng54321: Dropped the corresponding commit as this bug doesn't show now on my side either.

AiAe (approved):

PR #4108, opened 2024-05-08
Add periodic editor map backups

Emik03:
This PR adds a backup system for charting.

- Every 5 minutes, a scheduler checks if any modifications have been made to the chart, and in which case an automatic copy is saved in the form of `./Data/Backups/Charts/<NameOfQua>/<Timestamp>.qua`.
- Additionally, a check is added on the game's startup to delete every chart backup older than 48 hours.

The exact time intervals are chosen arbitrarily, although I don't think it's worth adding a setting for.


Emik03:
what

Emik03:
note for future self: do not rename branches if part of a PR, in fact don't rename branches in general

AiAe (approved):

PR #4111, opened 2024-05-09
[1.5.0] Add a toggle for deleting original file after it is imported

WilliamQiufeng:
Added a toggle so that you can keep the mapset files after importing them.

**TODO** Delete the file anyways if we are doing in-game download and import


WilliamQiufeng:
???

PR #4118, opened 2024-05-11
[1.6.0] Improve mapset download

WilliamQiufeng:
**Requires** https://github.com/Quaver/Quaver.Server.Client/pull/16 and https://github.com/Quaver/Quaver.Resources/pull/16

* [x] Adds the ability to stop, retry and remove downloads.
* [x] Retry only available after being stuck
* [x] Clear items after import
* [x] Fix bug: download blocking in queue (currently only checks IsDownloading but not if finished)8fa3/Quaver.Shared/Screens/Edit/Dialogs/EditorUploadingMapsetDialog.cs#L83))

It looks like this (subject to change):
image


WilliamQiufeng:
Should be ready for review

Swan (changes requested):
The download progress seems to break with this. When clicking on a mapset it goes from 0% to 100%, then stays there until it completes.

https://streamable.com/qoxoue

WilliamQiufeng:
> The download progress seems to break with this. When clicking on a mapset it goes from 0% to 100%, then stays there until it completes.
> 
> https://streamable.com/qoxoue

It seems like it is due to HEAD requests not being allowed in api-v2. AiAe has made relevant change for that (https://github.com/Quaver/api-v2/pull/119). Please check if it works.

Swan (approved):

PR #4125, opened 2024-05-14
[1.5.0] Add a search filter for scroll velocity changes

faluhub:
This pr adds a search filter for the amount of slider velocity changes in a map. Its short name is `svc`. Must refresh beatmaps before results show up :)

### DIFF
--- a/Quaver.Shared/Database/Maps/Map.cs
+++ b/Quaver.Shared/Database/Maps/Map.cs
@@ -214,6 +214,11 @@ namespace Quaver.Shared.Database.Maps
         public int OnlineOffset { get; set; }
 
         /// <summary>
+        ///     Returns the amount of times the scroll velocity changes in a map
+        /// </summary>
+        public int ScrollVelocityChanges { get; set; }
+
+        /// <summary>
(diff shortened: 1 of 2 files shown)


AiAe:
I think this needs to be improved to ignore scroll velocities at 1.00x, because they are marked falsely as scroll velocity. 
For example this map will show when searching for svc>0 https://quavergame.com/mapset/map/16

Emik03:
Wouldn't this also include osu! normalization SVs, so you can't just say "ignore 1x", it must be "ignore *perceived* 1x-ish". (rounded due to floating points)

Illuminati-CRAZ:
Clearly we should implement an SV difficulty calculator for this

Illuminati-CRAZ:
On a serious note, I personally don't think we should add scroll velocity _count_ as a thing to filter by. A bunch of "normal" maps will be would be filtered out by "svc = 0" because they use SVs for BPM normalization, and at higher sv counts it stops being a good metric for distinguishing between SV maps.

AiAe (changes requested):
Read PR comments

PR #4153, opened 2024-05-27
Attempt to fix score retrieving error when entering multiplayer gameplay

WilliamQiufeng:
Don't include players that we have not received info for when retrieving scores

This error might happen when: When the player retrieves the list, there happens to be an online user that we haven't retrieved an information for.


WilliamQiufeng:
superseded by #4159

PR #4154, opened 2024-05-28
Upgrade mapset filtering system

### LINKED ISSUE #4146: Upgrade mapset filtering system

Current search query uses regular expressions to match key-value pairs. It currently cannot handle searching for specifically titles (#4144), invert conditions (#4069) and other QoL features (#4006).

### Using a tokenizer and parser approach
This way, we could easily introduce new features into the search query:
* Specify length using a time span data type instead of only seconds (i.e. allow using things like `length>5m`)
* Include strings with spaces (`title: "aaa bb"`) (#4144)
* Use the Trie structure to automatically deduce the correct keywords from shorthands.
* Additional syntax:
  * *Includes* operator `:`: `title: Arro`. This should be introduced because `title=Arro` doesn't make sense. (#4144)
    * Optionally, the includes operator should use a fuzzy matching algorithm
  * Invert (not) operation, for example `!tag: sv` (#4069)
  * (potentially) OR operation using commas `,` and grouping of queries using brackets `()` (e.g. `d>30, d<20` or `!(d>20 d<30)` for maps with difficulty being either less than 20 or more than 20)
  * Allow some more shorthands (for example typing `unranked` directly means searching for `s=unranked`. Of course, these require full words to be typed, so terms like `unr` won't count)
* Care should be taken to make invalid search filters back to a normal text search. This means invalid queries like `length>30%` should be interpreted as searching for the literal text `length>30%`

### Add a panel under the search input box for visualized query builder
Adding a builder panel means the player doesn't need to have looked at the Wiki pages to learn all those keywords and syntaxes, and to know that the search filter even exists. This should also solve (#3937).  
The builder is responsible for generating the corresponding query when visualised items change, and updating the visual representation when the text box changes.
An example of such panel I can currently think of is the train departure condition layout in Factorio.
Locomotive_gui

### (Optional) Add a toggle for ascending/descending order of mapsets
This is a tiny feature regarding sorting order, so you could see, for example, least recently added maps, on top of the page. Of course, alternatively, you can either drag the scrollbar or use the middle mouse button to directly scroll to the bottom, so discussion is needed to decide if it's an overkill and a waste of screen space.

Regexes technically could be used to do all of the things above (except parenthesis matching since that's something not possible in normal regex but in CFGs. ~I won't bother doing [this](https://stackoverflow.com/questions/546433/regular-expression-to-match-balanced-parentheses)~), but I'm more inclined to make a dedicated parser for it. 

### Additional context

_No response_

QiuFeng54321:
Resolves #4146

It uses a simple parser I wrote which you can find [here](https://github.com/QiuFeng54321/WilliamQiufeng.SearchParser). It's referenced as a NuGet package.

Searching for maps whose creator contains `piano`:
image

Searching for maps whose creator doesn't contain `piano`, length less than 2 minutes, difficulty < 20:
image

**TODO**
+ [ ] Status and Game key doesn't work

### DIFF
--- a/Quaver.Shared/Database/Maps/MapsetHelper.cs
+++ b/Quaver.Shared/Database/Maps/MapsetHelper.cs
@@ -16,6 +16,8 @@ using Quaver.API.Helpers;
 using Quaver.Shared.Config;
 using Quaver.Shared.Database.Playlists;
 using Quaver.Shared.Modifiers;
+using WilliamQiufeng.SearchParser.Parsing;
+using WilliamQiufeng.SearchParser.Tokenizing;
 using Wobble.Bindables;
 
 namespace Quaver.Shared.Database.Maps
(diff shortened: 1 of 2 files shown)


Swan:
@WilliamQiufeng  Not sure if this is working as intended. None of the search filters are working anymore with this.

WilliamQiufeng:
> @WilliamQiufeng Not sure if this is working as intended. None of the search filters are working anymore with this.

Updated the library and repo and fixed several bugs including this.

Swan (approved):

PR #4045, opened 2024-04-12
[1.6.0] Add animation system for maps

WilliamQiufeng:
**Requires** https://github.com/Quaver/Quaver.API/pull/164 and https://github.com/Quaver/Wobble/pull/143

Allow using lua scripts to run an animation.

It currently has several ways of doing animations:
+ Segments and Triggers: A segment contains a function that is called repeatedly between a range of time. Triggers fire only once at a specific time.
   + A Tween Segment is added for smooth change in attributes of anything
   + An IntervalTrigger is used to fire a set number of times with a set interval
   + It supports dynamic adding and destroying, so you can use it with event triggers and other components
+ State Machines (Untested but added): A basic state machine that updates only the function of the current state. There can be multiple state machines at the same time.
+ Event trigger: You can set a lua callback for events like player hitting a note, or a new note entering the playfield.

Roadmap:
+ [x] Drawable Rotation
+ [x] Drawable scaling
  + [ ] Fix scaling orientation issue, observable with lane config preset
+ [x] Drawable inheritive tint and alpha
+ ~Game window manipulation~ Concluded as impossible
  + [ ] Letterboxing for wayland
+ [x] Shader support
+ [x] Layering system
+ [ ] Effect presets
  + [x] Lane swapping
  + [ ] Lane config (Changing the number of lanes e.g. 7K to 6K, etc.)
+ [x] Tweening scroll speed (#4210 )
+ [x] Beat-to-time converter (`beat(measure, beat` and `measure(measure)`)
+ [x] Audio sample loading and playing
+ [x] Script error count limiter
+ [x] Script update limiter
+ [x] Vectors
+ [x] Exec
+ [ ] Storage
+ [x] Fully index stage components
  + [x] LaneContainer(i) -> Lanes[i]
  + [x] UI
+ [ ] Design panel / Directive parsers

Check out https://github.com/WilliamQiufeng/QuaverAnimation for easy setup and example maps.

https://github.com/Quaver/Quaver/assets/32996262/83acbccb-c4fe-4117-95b1-6e837aaa92cb

https://github.com/Quaver/Quaver/assets/32996262/4836983b-39c0-4001-97c2-cc6d1a7e6720

https://github.com/Quaver/Quaver/assets/32996262/2da5705f-98d7-427d-8fd0-0b17d90484eb

https://github.com/Quaver/Quaver/assets/32996262/5b8aabd0-3b16-4730-b1ca-44cf68ce18cc

[...]

PR #4180, opened 2024-06-12
[1.6.0] Editor event system

### LINKED ISSUE #4051: Implement event handling to `state`

In my [plugin](https://github.com/ESV-Sweetplum/AFFINE), placing a group of timing lines and scroll velocities will add data into a table, which is then saved in a bookmark for future use. The purpose of this is to easily be able to delete or modify all of the placed timing lines and notes. However, problems arise when the user does an undo action; this does not undo any of the code, leading to a faulty tabular entry.

The syntax I'm thinking is similar to how web APIs are constructed:
```lua
state.on("undo", undoFunction)

function undoFunction()
-- execute code here
end
```
This syntax is very versatile and allows new features to be implemented very easily. Some ideas I have for emitted events and potential syntaxes would be the following:
```lua
state.on("load", loadFunction)
state.on("exit", exitFunction)
state.on("redo", redoFunction)
state.on("placeHitObject", function (hitObject) doSomething() end) 
state.on("removeHitObject"), function (hitObject) doSomething() end)
```

Of course, it's possible to create keybind wrappers using `imgui.IsKeyPressed()` and `imgui.IsKeyDown()`, like the following example to detect undos:
```lua
if (imgui.IsKeyDown(LeftControl) and imgui.IsKeyPressed(keys.Z)) then
  -- execute undo action
end
```
However, there are a couple issues with this:
- Other actions are possibly rebindable, and plugins are not exposed to the user's keybind settings
- `IsKeyPressed()` is quite unreliable with many plugin tabs open, as conflicts between the editor and the plugin happen very often. SV makers can testify that sometimes, it takes 2-3 key presses to execute an action that should only take one.
- Keybind wrappers are very limited, as there is no existing support for detecting clicks on a non-ImGui element.

Alternatively, we could detect current state and previously saved state and detect changes. In my AFFINE example, lineOffsets are actually saved. Technically, something like this could be possible:
```lua
function draw()
    imgui.Begin("Window")

    local previousLineCount = state.GetValue("lineCount") or 0
    local lineCount = #map.TimingPoints

    local desiredLineDifference = -5 -- Assume that the previous action of placing lines placed 5 of them.  Thus, we are looking for when 5 lines are deleted.

    if (lineCount - previousLineCount == desiredLineDifference) then 
         -- Since the number of lines that were deleted was equal to however 
         -- many lines was placed, assume an undo occurred
    end

    state.SetValue("lineCount", lineCount)

    imgui.End()
end
```
Hopefully it's easy to see why this isn't viable at a large scale.

### Additional context

_No response_

WilliamQiufeng:
Resolves #4051 with a few differences

This is a modified copy of the event system implemented in the animation pr (#4045):

Subscribe to events with `events.Subscribe(action_type.xxx, handler_function)`

Unsubscribe with `events.Unsubscribe`

`handler_function` has the form `function(instance: EventInstance)` where the `EventInstance` has fields:
* `EventInstance.Action` the action object. e.g. `instance.action.hitObject.startTime`
* `EventInstance.ActionType` type of the action (`action_type.xxx`)
* `EventInstance.IsUndo` indicates whether the action is performed(redid) or undid

The action type `action_type.None` is treated as a wildcard. If you subscribe a function to this type, it will be called if **any** action is performed or undid.

Example plugin code:
```lua
logs = ""
firstUpdate = true

function draw()
    if firstUpdate then
        firstUpdate = false
        -- Subscribes to any event
        events.subscribe(action_type.None, listener)
        -- Subscribes particularly to the placement of hitObjects
        events.subscribe(action_type.PlaceHitObject, notePlacementListener)
    end
    imgui.Begin("Event logger")
    imgui.Text(logs)
    imgui.End()
end

function addLog(s)
    logs = logs .. '\n' .. s
end

function listener(instance)
    addLog("Action captured by global: " .. tostring(instance.actionType) .. ", undo = " .. tostring(instance.isUndo))
end

function notePlacementListener(instance)
    addLog("Note place at " .. tostring(instance.action.hitObject.StartTime) .. ", undo = " .. tostring(instance.isUndo))
end
```

### DIFF
--- a/Quaver.Shared/Screens/Edit/Actions/Batch/EditorActionBatch.cs
+++ b/Quaver.Shared/Screens/Edit/Actions/Batch/EditorActionBatch.cs
@@ -13,7 +13,7 @@ namespace Quaver.Shared.Screens.Edit.Actions.Batch
 
         private EditorActionManager ActionManager { get; }
 
-        private List<IEditorAction> EditorActions { get; }
+        public List<IEditorAction> EditorActions { get; }
 
         [MoonSharpVisible(false)]
         public EditorActionBatch(EditorActionManager manager, List<IEditorAction> actions)
--- a/Quaver.Shared/Screens/Edit/Actions/Bookmarks/Add/EditorActionAddBookmark.cs
+++ b/Quaver.Shared/Screens/Edit/Actions/Bookmarks/Add/EditorActionAddBookmark.cs
@@ -15,7 +15,7 @@ namespace Quaver.Shared.Screens.Edit.Actions.Bookmarks.Add
 
         private Qua WorkingMap { get; }
 
-        private BookmarkInfo Bookmark { get; }
+        public BookmarkInfo Bookmark { get; }
 
         [MoonSharpVisible(false)]
         public EditorActionAddBookmark(EditorActionManager manager, Qua map, BookmarkInfo bookmark)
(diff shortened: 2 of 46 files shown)


WilliamQiufeng:
Closed since it has been implemented in #4105

PR #4181, opened 2024-06-13
Read quaver.cfg with UTF-8 encoding

WilliamQiufeng:
By default, ini is read with ASCII and written with UTF-8. This makes CJK characters not readable. This commit will fix it such that for example, the choice of an audio device with CJK characters in its name will be saved and loaded correctly, instead of falling back to Default. (My headphone is named '耳机 (Realtek Audio)' and it was read as '?????? (Realtek Audio)' which does not exist as a device, so it just falls back to Default)


AiAe:
Probably change PR to be to redesign branch. 
Also is this a breaking change to current config files? Will they break/be unable to be read properly?

PR #4211, opened 2024-07-10
[1.6.0] Add hit bubbles

WilliamQiufeng:
+ ~~Config for tourneys~~ hard

Testing skin:  (Please rename to .qs)
[BarForUse.zip](https://github.com/user-attachments/files/17685299/BarForUse.zip)

You can tweak the values in the ini:
* The HitBubblesRecordedJudgements is the acceptable judgements that are shown as hit bubbles. It can take any `Judgement`s, like Marv, Perf, or a list of them (like `Perf, Okay, Miss`) or `NoMarv` for everything not Marv, `All` for every judgement, `Default` for default setting (`NoMarv`)
* HitBubblesAlignment can be `LeftStage`, `RightStage` or `BelowStage`, which means the docking position of the bubbles
* HitBubblesType is the falling direction of the bubbles. It can be `FallUp`, `FallDown`, `FallLeft` or `FallRight`


Swan (changes requested):
Looks good. It doesn't seem to work with the Autoplay mod on though, so we should take care of that before merging.

WilliamQiufeng:
> Maybe the default should show all judgements. Any reason you chose differently?

I think it is because that in general marvelous judgements will be spamming the hit bubbles, so other kinds of judgements disappear very fast. In tourneys it should be intended to be used so that the commentators can see recent judgements in the past seconds so they don't miss anything.

Swan:
> > Maybe the default should show all judgements. Any reason you chose differently?
> 
> I think it is because that in general marvelous judgements will be spamming the hit bubbles, so other kinds of judgements disappear very fast. In tourneys it should be intended to be used so that the commentators can see recent judgements in the past seconds so they don't miss anything.

How about we always display all judgements on autoplay then, but during gameplay it can be NoMarv?

WilliamQiufeng:
> How about we always display all judgements on autoplay then, but during gameplay it can be NoMarv?
Done

Quaver.Shared/Screens/Gameplay/UI/Bubble/HitBubbles.cs
```diff
@@ -20,31 +20,41 @@ public class HitBubbles : Container
...
+    private bool HideHitBubbles =>
+        (Screen?.IsCalibratingOffset ?? false)
+        || SkinKeys.HitBubblesScale == 0
+        || !ConfigManager.DisplayHitBubbles.Value;
+
+    private HitBubbleRecordedJudgement RecordedJudgements =>
+        Screen.InReplayMode
+            ? HitBubbleRecordedJudgement.All
+            : SkinKeys.HitBubblesRecordedJudgements;
+
```
Swan, line 36: Maybe check if the Autoplay mod is active here instead of replay mode, since we don't want to display all in tournaments or regular replays.

Swan (changes requested):

Quaver.Shared/Screens/Gameplay/UI/Bubble/HitBubbles.cs
WilliamQiufeng: Changed the logic. It still has replay mode check but with constraint that we are playtesting, since I still want All to apply when we are previewing.

Swan (approved):

PR #4212, opened 2024-07-10
[1.5.0] Add a toggle for deleting original file after it is imported

WilliamQiufeng:
#4111 I do mess up PRs


AiAe (changes requested):
Whats up with this PR and branch, why does it include canary submodules?

Warp9000:
looks like its based on develop branch

PR #4219, opened 2024-07-15
[1.5.0] Cap FPS to 500 if FPS mode is set to Custom and the custom fps is set above 500

WilliamQiufeng:
~~Side effect: If the editor crashes, the Custom FPS limit might not be recovered. I don't know if I can do anything with that.~~ **EDIT** No it will work properly now.

### DIFF
--- a/Quaver.Shared/Screens/Edit/EditScreenView.cs
+++ b/Quaver.Shared/Screens/Edit/EditScreenView.cs
@@ -4,6 +4,7 @@ using Microsoft.Xna.Framework;
 using Quaver.API.Enums;
 using Quaver.API.Maps;
 using Quaver.Shared.Assets;
+using Quaver.Shared.Config;
 using Quaver.Shared.Graphics.Backgrounds;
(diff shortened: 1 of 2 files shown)


Quaver.Shared/Screens/Edit/EditScreenView.cs
```diff
@@ -91,12 +92,33 @@ public class EditScreenView : ScreenView
...
+        ///     Maximum allowed custom FPS in editor
+        /// </summary>
+        private const int MaximumCustomFps = 500;
+
+        /// <summary>
+        ///     If the fps is capped, the previous custom fps value.
+        ///     Otherwise, it will be 0
+        ///     When the screen is destroyed, the FPS cap should return to <see cref="previousCustomFps"/>
+        /// </summary>
+        private int previousCustomFps;
```
Warp9000, line 105: use `ConfigManager.CustomFpsLimit.Value` instead

AiAe (approved):

Warp9000 (approved):

PR #4221, opened 2024-07-15
[1.5.0] Allow placing long notes during livemapping

### LINKED ISSUE #4220: Rebindable/gameplay keys for live mapping + addition of LNs to live mapping

I would love to do some live mapping, but comfortably using my binds instead of the numrow. 
The addition of LNs would also be huge. I do feel like there is potential to at least get a base for a map + timing point info by playing to a song you know, but in the current state it's hard to test it properly. Would be amazing if we could better utilize the live mapping feature. 

- Addition of either enabling gameplay binds for livemapping, or specific separate keybinds
- Addition of the ability to hold down keys while livemapping to create a longnote.

_No response_

### Additional context

_No response_

WilliamQiufeng:
Resolves #4220

Adds two menu options: a toggle for placing long notes during livemapping (defaults to true), and a setting for the minimum threshold to hold the key to recognize it as a long note (defaults to 100ms).

Difficulty bar recreation creation is now debounced (200ms), because it keeps getting called when resizing long notes, which just eats the entire memory and fps.

Keybind is not added yet; I'm thinking maybe we can do that when we move everything to the new unused input manager.

### DIFF
--- a/Quaver.Shared/Screens/Edit/Dialogs/EditorSetLiveMapLongNoteThresholdDialog.cs
+++ b/Quaver.Shared/Screens/Edit/Dialogs/EditorSetLiveMapLongNoteThresholdDialog.cs
@@ -1,0 +1,79 @@
+using System;
+using System.Text.RegularExpressions;
+using Quaver.Shared.Assets;
+using Quaver.Shared.Config;
+using Quaver.Shared.Graphics;
+using Quaver.Shared.Helpers;
+using Wobble.Graphics;
+using Wobble.Graphics.UI.Form;
+using Wobble.Managers;
+
+namespace Quaver.Shared.Screens.Edit.Dialogs
+{
+    public class EditorSetLiveMapLongNoteThresholdDialog : YesNoDialog
+    {
+        private EditScreen Screen { get; }
+
+        /// <summary>
+        /// </summary>
+        protected Textbox Textbox { get; set; }
+
+        /// <inheritdoc />
+        /// <summary>
+        /// </summary>
+        public EditorSetLiveMapLongNoteThresholdDialog(EditScreen screen) : base("SET LIVEMAP LN THRESHOLD",
+            "Enter the minimum length of a long note during live-mapping...")
+        {
+            Screen = screen;
+            CreateTextbox();
+
+            Panel.Height += 50;
+            YesButton.Y = -30;
+            NoButton.Y = YesButton.Y;
+
+            YesButton.Clicked += (sender, args) => OnSubmit(Textbox.RawText);
+        }
+
+        /// <summary>
+        /// </summary>
+        private void CreateTextbox()
+        {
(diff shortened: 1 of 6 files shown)


AiAe:
The issue also mentions they want to rebind the default key binds for live mapping from 1 2 3 4. 
This PR does not fully resolve the issue. 
Is this planned for another PR?

WilliamQiufeng:
> Is this planned for another PR?

It is. I am thinking about moving them altogether to the new editor input manager

AiAe:
> > Is this planned for another PR?
> 
> It is. I am thinking about moving them altogether to the new editor input manager

I see, I assume you will have to open/fork Ice's old PR to implement his idea/vision?

WilliamQiufeng:
> I see, I assume you will have to open/fork Ice's old PR to implement his idea/vision?

I think so. If thngs haven't changed much, I want to continue it from where it was left off some time. BTW I'm also hoping to move the config file to that structured config I added for editor window saving.

Quaver.Shared/Screens/Edit/EditScreen.cs
```diff
@@ -873,7 +880,6 @@ private void HandleKeyPressDelete()
 
             DeleteSelectedObjects();
         }
-
```
Warp9000, line 876: this empty line should stay

Quaver.Shared/Screens/Edit/Dialogs/EditorSetLiveMapLongNoteThresholdDialog.cs
```diff
@@ -0,0 +1,79 @@
...
+                FontManager.GetWobbleFont(Fonts.LatoBlack),
+                20, ConfigManager.EditorLiveMapLongNoteThreshold.Value.ToString(),
+                "Enter an threshold of LNs during livemapping...", OnSubmit)
+            {
+                Parent = Panel,
+                Alignment = Alignment.BotCenter,
+                Y = -100,
+                Tint = ColorHelper.HexToColor("#2F2F2F"),
+                AlwaysFocused = true,
+                AllowedCharacters = new Regex(@"^[0-9-]*$")
```
Warp9000, line 51: this regex should probably be `^[0-9]*$`

Quaver.Shared/Screens/Edit/Dialogs/EditorSetLiveMapLongNoteThresholdDialog.cs
```diff
@@ -0,0 +1,79 @@
...
+        {
+            try
+            {
+                var val = int.Parse(s);
+                ConfigManager.EditorLiveMapLongNoteThreshold.Value = val;
+            }
+            catch (Exception)
+            {
+                // ignored
+            }
```
Warp9000, line 76: TryParse should probably be used instead

Warp9000 (changes requested):

Warp9000 (approved):

PR #4233, opened 2024-07-19
[1.6.0] Add keybind actions for new keybind system

WilliamQiufeng:
Supercedes #3725 
Requires https://github.com/Quaver/Wobble/pull/152

This PR adds all missing actions while modifying and adding the following:
* Default keybind: `Home` and `End` should have the same `+select` semantics when with modifier.
* Default keybind: `Alt+D#` is not used for changing snaps. Instead, they are used to swap lanes of selected hit objects
* Mouse buttons & Mouse scrolling is added in the Wobble PR described above. `M_Up` and `M_Down` for scrolling through time, `Ctrl+M_Up/M_Down` for changing snaps. Scrolling can be inverted via editor invert scrolling option.
* Adds a `Keybinds` menu section as described in the original PR. You can reset keybinds, fill missing keybinds and change scrolling inversion settings there.
* The existing `ConfigManager.InvertEditorScrolling` is used instead of `EditorInputConfig.ReverseScrollSeekDirection`. Toggles for scrolling inversion (scrolling and beat snap changing) are moved to the `Keybinds` menu section


Swan (changes requested):
When livemapping, you can place another note within an LN. It's very easy to create overlaps.

If you place a note inside an existing LN, I think it should either delete the LN and replace.

Not sure if this is issue already existed, but if we're already here we should fix it.

WilliamQiufeng:
> When livemapping, you can place another note within an LN. It's very easy to create overlaps.
> 
> If you place a note inside an existing LN, I think it should delete the LN and replace.

Done. It also considers removing the notes that are within the livemapped LN now.

I changed the logic of LN placement so it only transforms the note into an LN when we release the key. This is because repeatedly calling ResizeLongNote spams the undo/redo stack.

I also made undo/redo repeatable.

Swan (changes requested):
Placing LNs with the keyboard seems to have regressed with the new changes.

WilliamQiufeng:
> I changed the logic of LN placement so it only transforms the note into an LN when we release the key. This is because repeatedly calling ResizeLongNote spams the undo/redo stack.

You can now only see the note becoming LNs when you release keys. I did this so there is no spamming of ResizeLongNote

Swan:
> > I changed the logic of LN placement so it only transforms the note into an LN when we release the key. This is because repeatedly calling ResizeLongNote spams the undo/redo stack.
> 
> You can now only see the note becoming LNs when you release keys. I did this so there is no spamming of ResizeLongNote

I'm not seeing the LNs appear even after releasing. There's also an intense flashing with the difficulty bar

See the video below
https://streamable.com/ypklsw

WilliamQiufeng:
> I'm not seeing the LNs appear even after releasing. There's also an intense flashing with the difficulty bar

My bad. should be fixed now.

Swan (approved):

PR #4286, opened 2024-09-12
Migrate API usage to v2

Swan:
Everything except the scoreboard and playlists will use api-v2.

A separate PR will come for the scoreboard in the future. It requires a very large rewrite.


EpiCreeper:
Playlists kinda suck currently. I get it isn't top priority but they are pretty limited wothout the difficulties getting groupped into mapsets

Swan:
@EpiCreeper Please refrain from making comments that don't add to the topic at hand. This has nothing to do with playlists and their functionality.

EpiCreeper:
Would you be able to update playlists behavior without updating them to the v2 api? Sorry for being unclear but thats what i was worried about in my comment earlier.

PR #4300, opened 2024-10-11
Load the background image when instantiating `EditScreenView`.

### LINKED ISSUE #4114: map bg doesn't change when switching difficulties in the editor
### Describe the bug

on a map with different background for different difficulties, the background doesn't change when switching difficulties using the editor

### To Reproduce

1. edit a map with multiple difficulties which also has different bg
2. file -> switch difficulty
3. the background stays on the first opened map

### Expected behavior

the background should change to the switched difficulty bg rather than staying the same

### Log Files

-

### Screenshots

_No response_

lanylow:
Checks if the map's background image has changed and loads the new one if necessary. Fixes #4114.

### DIFF
--- a/Quaver.Shared/Screens/Edit/EditScreenView.cs
+++ b/Quaver.Shared/Screens/Edit/EditScreenView.cs
@@ -5,10 +5,12 @@ using Quaver.API.Enums;
 using Quaver.API.Maps;
 using Quaver.Shared.Assets;
 using Quaver.Shared.Config;
+using Quaver.Shared.Database.Maps;
 using Quaver.Shared.Graphics.Backgrounds;
 using Quaver.Shared.Graphics.Graphs;
 using Quaver.Shared.Graphics.Menu.Border;
 using Quaver.Shared.Helpers;
+using Quaver.Shared.Scheduling;
 using Quaver.Shared.Screens.Edit.UI.AutoMods;
 using Quaver.Shared.Screens.Edit.UI.Footer;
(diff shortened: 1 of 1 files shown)


Emik03 (approved):

Quaver.Shared/Screens/Edit/EditScreenView.cs
```diff
@@ -131,6 +133,12 @@ public EditScreenView(Screen screen) : base(screen)
             EditScreen.BackgroundBrightness.ValueChanged += OnBackgroundBrightnessChanged;
             BackgroundHelper.Loaded += OnBackgroundLoaded;
             Footer.Parent = Container;
+            
+            ThreadScheduler.Run(() =>
```
Swan, line 137: You don't need to run this on another thread. The Load method already does this and has cancellation logic.

Swan (changes requested):

Quaver.Shared/Screens/Edit/EditScreenView.cs
lanylow: I wasn't sure whether to create a thread or not, but I decided to do it because of the comment written [here](https://github.com/Quaver/Quaver/blob/develop/Quaver.Shared/Screens/Selection/UI/FilterPanel/SelectFilterPanel.cs#L342). If you can confirm that it's not needed in this case, I will remove it.

Swan: I see. I think it's fine then.

Swan (approved):

PR #4380, opened 2024-11-25
Allow plugin to set selected scroll group id

WilliamQiufeng:
added setter to `state.SelectedScrollGroupId` so

```lua
state.SelectedScrollGroupId = "abc"
```

Can be used to select a different scrollgroup


ESV-Sweetplum:
big double u

AiAe (approved):

PR #4392, opened 2024-11-29
Update NoteVisualizer.cs (Fixes #4391)

### LINKED ISSUE #4391: NoteVisualizer in the Main Menu scrolls Inconstantly (Duplicate of #3384)
### Describe the bug

(Duplicate of https://github.com/Quaver/Quaver/issues/3384)
The old post was marked as completed, despite not having any changes in the code or ingame.

`NoteVisualizer.cs` has a problem where the first instance of the `note-visualizer` sprite scrolls slightly faster than any of the clones it creates afterwards.

### To Reproduce

1. Open game
2. Wait

### Expected behavior

(in Quaver/Quaver.Shared/Screens/Main/UI/Visualizer/NoteVisualizer.cs/)

To quote from the previous post:
"The first time the menu sprite is made, it just has a path to the bottom of the screen. But when it's cloned, it arbitrability adds 100 from its position, thus making the path it takes slightly slower. This becomes a problem when the `note-visualizer` texture is replaced to appear seamless."

This would be due to its movement being based on an easing function from its current position. (Code within `CreateSprites()`)
`sprite.MoveToY((int) WindowManager.Height, Easing.Linear, ANIM_TIME);`

When it has decided that the sprite has reached the bottom of the screen. It uses the `CreateSprites()` function from its current position, with the added arbitrary amount.
`CreateSprites((int) -firstSprite.Height + 100)`

This is what causes the visual bug in the first place. I don't know if I was clear enough the first time, I apologize.

### Log Files

bruh.
[runtime.log](https://github.com/user-attachments/files/17956223/runtime.log)
[network.log](https://github.com/user-attachments/files/17956233/network.log)

### Screenshots

[Video from old post](https://user-images.githubusercontent.com/66524513/152632099-ceca0c6f-3631-4fc4-9387-4498d27c1776.mp4)
[New Video](https://github.com/user-attachments/assets/e7141f40-f2d6-48b1-bee7-5534a88001bd)
Texture was modified to show inconsistencies.

Mr-Turnip:
Fix for inconsistent moving speed of note-visualizer texture

### DIFF
--- a/Quaver.Shared/Screens/Main/UI/Visualizer/NoteVisualizer.cs
+++ b/Quaver.Shared/Screens/Main/UI/Visualizer/NoteVisualizer.cs
@@ -67,7 +67,7 @@ namespace Quaver.Shared.Screens.Main.UI.Visualizer
             var firstSprite = Sprites.First();
 
             if (firstSprite.Y >= 0 && Sprites.Count == 1)
-                CreateSprites((int) -firstSprite.Height + 100);
+                CreateSprites((int) -firstSprite.Height);
 
             if (!RectangleF.Intersects(firstSprite.ScreenRectangle, new RectangleF(0, 0, WindowManager.Width, WindowManager.Height))
                 && firstSprite.Y > 0)


Mr-Turnip:
how tf do i do this?

AiAe:
This change will also make the default arrow animation start really "late" because it will be moved backwards. 
Unlike your skin ours have more padding.
I don't think this would be a change i'd merge but will leave open if maybe Swan decides it would be good idea.

PR #4399, opened 2024-12-12
[QoL] Add option to scale scroll speed with rate

WilliamQiufeng:
This PR adds an option to scale scroll speed with the rate applied, so that SV maps does not need manual scroll speed change to be comfortably read when applied rates.

The new option is a slider of percentage strength. The perceived scroll speed will be the set scroll speed multiplied by $1 + (rate - 1) * \frac{strength}{100}$, so that 0 strength is the default behaviour, 100 strength means the spacing between notes will be the same no matter the rate, and 50 means something in-between: the spacing will change, but not as drastic as 0 and not as static as 100.

Care is taken to constraint the perceived scroll speed between 5 and 100 (which is the original bound), so that SV maps with 100x SV as teleport won't accidentally show its notes early.

Also, the submitted replay will record the perceived scroll speed instead of the set scroll speed.

Here is an example: 

https://github.com/user-attachments/assets/f3d275bd-d1a1-4614-b582-b906e7e9c6f1

Notice that when the strength is set to 100, the spacing won't change when the rates change. This is very convenient for practising SV maps:
+ By making the spacing (near) constant, you can start practicing from a lower rate and slowly go back to 1.0x and the experience would be more familiar, as the notes you see don't change places.
+ When practising SV maps with rates, normally you really need this adjustment to at least make it readable (for example, when practising 1.2x you normally need to add a few scroll speed so it looks better) . This automates the need to manually adjust scroll speed.

### DIFF
--- a/Quaver.Shared/Config/ConfigManager.cs
+++ b/Quaver.Shared/Config/ConfigManager.cs
@@ -214,6 +214,11 @@ namespace Quaver.Shared.Config
         internal static BindableInt ScrollSpeed7K { get; private set; }
 
         /// <summary>
+        ///     Percentage of scaling applied when changing rates
+        /// </summary>
+        internal static BindableInt ScaleScrollSpeedStrengthPercentage { get; private set; }
+
+        /// <summary>
         ///     Direction in which hit objects will be moving for 4K gamemode
         /// </summary>
         internal static Bindable<ScrollDirection> ScrollDirection4K { get; private set; }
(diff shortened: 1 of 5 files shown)


EpiCreeper:
Ok cool now make it negative

WilliamQiufeng:
Trust me you never want negative values 

https://github.com/user-attachments/assets/e1dc28f0-1134-4879-b041-cb186aedd8a0

Emik03:
Now that I think about it, this option is only useful in SV charts. Should you change it to only apply to those or maybe add a toggle? Remember that the definition of an SV chart should consider whether the SVs are there to normalize BPMs.

WilliamQiufeng:
> ... whether the SVs are there to normalize BPMs.

Sorry I might need more explanation on what this means. 

There was a discussion around detecting SV maps in #4125. In gameplay we always use the map where SV is normalised, should the `BPM affects SV` be on. With that, would it suffice just to test if there are SV points with multipliers outside of a range in the normalised map?

Emik03:
Correct, my personal definition of SV is that at any point the scroll falls outside of the range 0.99-1.01. I do consider a map with constant 2x SV to be an SV map, but if you disagree you are free to make it relative to the first SV as well.

AiAe (approved):

PR #4443, opened 2025-03-13
Add toggle to display ranked accuracy on local leaderboards

BakersDozenBagels:
Adds a configuration setting to replace accuracy for the `Local` leaderboard during song select with ranked accuracy (which all other leaderboards do by default). This makes switching judgement windows less awkward, since you can more easily see your own growth and determine a target (ranked) accuracy to aim for.

I considered merging this with `ConfigManager.DisplayRankedAccuracy`, but I did not since comparing scores and setting score are separate activities with separate requirements (e.g. I personally do not set `DisplayRankedAccuracy`, but would set this).

## Example
The higher PR play here uses `Extreme`, while the lower PR play uses `Standard*`.
<details>
    <summary>Disabled</summary>
    
</details>
<details>
    <summary>Enabled</summary>
    
</details>

### DIFF
--- a/Quaver.Shared/Config/ConfigManager.cs
+++ b/Quaver.Shared/Config/ConfigManager.cs
@@ -274,6 +274,11 @@ namespace Quaver.Shared.Config
         internal static Bindable<bool> DisplayRankedAccuracy { get; private set; }
 
         /// <summary>
+        ///     Display the ranked accuracy for local leaderboards instead of the custom judgement windows accuracy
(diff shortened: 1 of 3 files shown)


AiAe:
@BakersDozenBagels can you remove all the reformatting you have pushed with this code? It does not appear to follow our standard I think + will make the code cleaner to review.

AiAe (approved):

PR #4470, opened 2025-05-23
Add mines

WilliamQiufeng:
Requires https://github.com/Quaver/Quaver.API/pull/200
+ [ ] Requires a resource change

This PR adds visual and judgement code to mine notes introduced in the relevant Quaver.API PR. In the video below, mine notes are shown as red hollow rectangle notes.
When you happen to be pressing the lane between the marvelous window of the mine you will get a miss, otherwise a marvelous hit is generated

https://github.com/user-attachments/assets/4b5d1c34-fd71-4bae-830d-bd3cef1dd405

hehe


ESV-Sweetplum:
I'm fine with this so long as all hitobjects are checked within range of receptors before the mines (as in prioritize a worse judgement over hitting a mine)

WilliamQiufeng:
> I'm fine with this so long as all hitobjects are checked within range of receptors before the mines (as in prioritize a worse judgement over hitting a mine)

Wait could you give an example? If I get it right wouldn't that mean that it could make the judgement window of the mine extremely small, or even zero, by putting a note right next to it?

ESV-Sweetplum:
Suppose at state.SongTime 200ms, there is a mine at 195ms and a note at 250ms, all in the same column. If an input is pressed, the note should take priority, even though the mine is closer to the receptor. This means that the player will receive a great judgement, and the mine will go by unnoticed.

ESV-Sweetplum:
> > I'm fine with this so long as all hitobjects are checked within range of receptors before the mines (as in prioritize a worse judgement over hitting a mine)
> 
> Wait could you give an example? If I get it right wouldn't that mean that it could make the judgement window of the mine extremely small, or even zero, by putting a note right next to it?

I suppose that's true, but personally I think it's on the mapper to avoid spamming mines such that this happens often

7ef38c:
> Suppose at state.SongTime 200ms, there is a mine at 195ms and a note at 250ms, all in the same column. If an input is pressed, the note should take priority, even though the mine is closer to the receptor. This means that the player will receive a great judgement, and the mine will go by unnoticed.

This implementation would make minedodge trivial no? While sure it is a more niche gimmick and either way,I don't think a player should just be able to ignore a mine if it would be in the note's great window, regardless of gimmicks.

ESV-Sweetplum:
maybe but i can only imagine the frustration if there's a mine like just a little bit before a note, and if you hit slightly early instead of a great you just shitmiss

edit: i'm also interested if `HitObjectType` is an exposed global as I don't see implementation for it in the files changed but i might just be tripping

Storn42:
> maybe but i can only imagine the frustration if there's a mine like just a little bit before a note, and if you hit slightly early instead of a great you just shitmiss
> 
> edit: i'm also interested if `HitObjectType` is an exposed global as I don't see implementation for it in the files changed but i might just be tripping

whats stopping people from making a map with 1ms holds, or stacked notes, or a constant 100x sv? I dont think making a gameplay element trivial just because its sometimes annoying is the best way to go. some people find SV and LN frustrating to play as well, should we make LN tails count as long as you hold them until the marv window? I'm not saying mines need to have a massive hit window, but changing gameplay to make mine behavior to basically not matter isn't the way to go.

WilliamQiufeng:
> edit: i'm also interested if `HitObjectType` is an exposed global as I don't see implementation for it in the files changed but i might just be tripping

HitObjectType is declared in the Quaver.API PR (https://github.com/Quaver/Quaver.API/blob/0e52858d364f5e8a53f1311ccdc679b8c31985fb/Quaver.API/Enums/HitObjectType.cs)

ESV-Sweetplum:
> > maybe but i can only imagine the frustration if there's a mine like just a little bit before a note, and if you hit slightly early instead of a great you just shitmiss
> > edit: i'm also interested if `HitObjectType` is an exposed global as I don't see implementation for it in the files changed but i might just be tripping
> 
> whats stopping people from making a map with 1ms holds, or stacked notes, or a constant 100x sv? I dont think making a gameplay element trivial just because its sometimes annoying is the best way to go. some people find SV and LN frustrating to play as well, should we make LN tails count as long as you hold them until the marv window? I'm not saying mines need to have a massive hit window, but changing gameplay to make mine behavior to basically not matter isn't the way to go.

that's true, i was under the assumption that mines would have an identical hit window to normal notes. if that's not the case then it should be fine

Quaver.Shared/Database/Maps/Map.cs
```diff
@@ -186,14 +187,19 @@ public class Map
         /// </summary>
         public int LongNoteCount { get; set; }
 
+        /// <summary>
+        ///     The count of long notes.
+        /// </summary>
+        public int MineCount { get; set; }
```
menvae, line 193: typo in docs

AiAe:
@codex review

AiAe (approved):

Warp9000 (approved):

Issue #4290, opened 2024-09-19
The "No Release Judgement" Modifier


EpiCreeper:
### Is your feature request related to a problem? Please describe.

I personally really like the way LNs are implemented in quaver, but most rhythm games don't require you to release the note in time. Some players transition from other games and find this change challenging and uncomfortable (and others are just bad at the game _cough cough_). These players are likely to turn on NLN, but some maps rely on the long notes to be coherent and interesting, which means NLN players are missing out.

### Describe the solution you'd like

Add a new modifier called No Release Judgement (NRJ) that would change the LN input like so: miss if you released early, marv if you held all the way, even if you didn't release in time. The difficulty of the map would be recalculated as if the NLN modifier is applied.

### Describe alternatives you've considered

So far it seems you dont wanna make any significant changes to the core gameplay, but you could potentially add a new type of LN without release timing, so that mappers could add flavor with LNs without ruining scores. In that case the NRJ mod would just replace al the LNs with the new kind.

### Additional context

I just want people to enjoy LN without being intimidated by the difficulty of them. With this modifier you won't quite get the same experience but I think more people would come to appreciate LN mapping and the effort behind it.

wardahub-design:
download
https://bit.ly/3XQLHlH
If you don't have the c compliator, install it.(gcc or clang)

ESV-Sweetplum:
This would be fantastic.

Issue #4271, opened 2024-08-24
Custom timing windows: Customizable "held note not released" judgement


wisheth:
### Is your feature request related to a problem? Please describe.

When I use a very draconic timing window preset I've made without goods, not releasing a hold note leads to a "good" rating that doesn't make sense in the context.

### Describe the solution you'd like

I'd like the rating assigned when not releasing a hold note to be customizable (e.g. changeable to a standard miss, or any other judgement). This could either just be a visual change or fully change the judgement logic depending on how difficult the latter is to implement, as I'm not sure how this case is calculated score wise.

### Describe alternatives you've considered

It's hard to find alternatives here besides "grin and bear it". I imagine an alternative visually would be to change one of the existing judgements in this to a good, but that feels very dishonest and would skew scoring in a way that's not necessarily predictable when directly compared to what I currently have.

PR #4481, opened 2025-06-19
Customizable LN Missed Release Judgements

BakersDozenBagels:
Requires https://github.com/Quaver/Quaver.API/pull/202

Implements https://github.com/Quaver/Quaver/issues/4271, and should be usable for https://github.com/Quaver/Quaver/issues/4290 by setting it to Marvelous.

https://github.com/user-attachments/assets/ce9621df-2d2d-48e9-af7a-aacd20bd1cdd

Releasing an LN late will not give a worse judgement than never releasing it.
Additionally, if the judgement is set to Okay or Miss (i.e. stricter than default), it becomes possible to get an Okay by releasing within the relevant window.

### DIFF
--- a/Quaver.Shared/Database/Judgements/JudgementWindowsDatabaseCache.cs
+++ b/Quaver.Shared/Database/Judgements/JudgementWindowsDatabaseCache.cs
@@ -1,10 +1,9 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using Quaver.API.Enums;
 using Quaver.API.Maps.Processors.Scoring;
 using Quaver.Shared.Config;
-using Quaver.Shared.Database.Settings;
-using SQLite;
 using Wobble.Bindables;
 using Wobble.Logging;
 
@@ -95,7 +94,7 @@ namespace Quaver.Shared.Database.Judgements
         {
             var windows = DatabaseManager.Connection.Table<JudgementWindows>().ToList();
             foreach (var w in windows)
-                w.LNMissJudgement ??= API.Enums.Judgement.Good;
+                w.LNMissJudgement ??= Judgement.Good;
             return windows;
         }
 
--- a/Quaver.Shared/Database/Judgements/JudgementWindowsDatabaseCache.cs
+++ b/Quaver.Shared/Database/Judgements/JudgementWindowsDatabaseCache.cs
@@ -93,7 +93,10 @@ namespace Quaver.Shared.Database.Judgements
         /// </summary>
         private static List<JudgementWindows> FetchAllWindows()
         {
-            return DatabaseManager.Connection.Table<JudgementWindows>().ToList();
+            var windows = DatabaseManager.Connection.Table<JudgementWindows>().ToList();
+            foreach (var w in windows)
+                w.LNMissJudgement ??= API.Enums.Judgement.Good;
+            return windows;
         }
 
         /// <summary>
--- a/Quaver.Shared/Screens/Selection/UI/Modifiers/Dialogs/Windows/JudgementWindowComboBreakDropdown.cs
+++ b/Quaver.Shared/Screens/Selection/UI/Modifiers/Dialogs/Windows/JudgementWindowComboBreakDropdown.cs
@@ -1,17 +1,29 @@
 ﻿using Quaver.API.Enums;
 using Quaver.API.Maps.Processors.Scoring;
+using System.Collections.Generic;
 using Wobble.Bindables;
 
 namespace Quaver.Shared.Screens.Selection.UI.Modifiers.Dialogs.Windows
 {
     public class JudgementWindowComboBreakDropdown : JudgementWindowDropdown
     {
+        // This dropdown excludes Marvelous (`(Judgement)0`), since that would entirely disable combo.
+        // This means we need to offset the dropdown value by one.
         protected override int Target
         {
             get => (int)SelectedWindow.Value.ComboBreakJudgement - 1;
             set => SelectedWindow.Value.ComboBreakJudgement = (Judgement)(value + 1);
         }
 
-        public JudgementWindowComboBreakDropdown(Bindable<JudgementWindows> selectedWindow) : base(selectedWindow, "Combo Break Judgement: ") { }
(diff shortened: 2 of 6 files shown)


Warp9000 (changes requested):
existing presets gets set to marvelous, they should be set to good.

the combo break dropdown gets covered by the new dropdown
image

BakersDozenBagels:
@Warp9000 Both issues should now be fixed.

Quaver.Shared/Database/Judgements/JudgementWindowsDatabaseCache.cs
```diff
@@ -93,7 +93,10 @@ private static void InitializeDefaultPresetWindows()
         /// </summary>
         private static List<JudgementWindows> FetchAllWindows()
         {
-            return DatabaseManager.Connection.Table<JudgementWindows>().ToList();
+            var windows = DatabaseManager.Connection.Table<JudgementWindows>().ToList();
+            foreach (var w in windows)
+                w.LNMissJudgement ??= API.Enums.Judgement.Good;
```
WilliamQiufeng: I feel like you could import `Quaver.API.Enums` and say `Judgement.Good`

Quaver.Shared/Screens/Selection/UI/Modifiers/Dialogs/Windows/JudgementWindowComboBreakDropdown.cs
```diff
@@ -1,62 +1,17 @@
...
-        {
-            if (SelectedWindow.Value.IsDefault && Dropdown.Opened)
-                Dropdown.Close();
-
-            Dropdown.IsClickable = !SelectedWindow.Value.IsDefault;
-            base.Update(gameTime);
-        }
-
-        public override void Destroy()
+        protected override int Target
```
WilliamQiufeng, line 12: Please document this property to show the reason for subtracting/adding 1.

Quaver.Shared/Screens/Selection/UI/Modifiers/Dialogs/Windows/JudgementWindowDropdown.cs
```diff
@@ -0,0 +1,63 @@
...
+
+        private static List<string> GetDropdownItems() => new()
+            {
+                "Perfect",
+                "Great",
+                "Good",
+                "Okay",
+                "Miss"
+            };
+
```
WilliamQiufeng: You provided a default dropdown items for the abstract `JudgementWindowDropdown`, but it feels odd that the default options are missing `Marlevous`. You could make `List<string> options = null` a **required** field in the constructor, and move this `GetDropdownItems()` up to `JudgementWIndowComboBreakDropdown`, as it can annotate better for your custom `Target` getter and setter (in the same file too).
In another word, `GetDropdownItems()` should be implemented separately in each child class of this abstract class.

BakersDozenBagels: It might make sense to make every judgement the default. However, YAGNI, so I'll just clean up the organization like you suggest.

BakersDozenBagels: Resolved in 1dd6db7

Quaver.Shared/Screens/Selection/UI/Modifiers/Dialogs/Windows/JudgementWindowComboBreakDropdown.cs
BakersDozenBagels: Resolved in 1dd6db7

Quaver.Shared/Database/Judgements/JudgementWindowsDatabaseCache.cs
BakersDozenBagels: Resolved in 5f90b22

Quaver.Shared/Screens/Selection/UI/Modifiers/Dialogs/Windows/JudgementWindowDialog.cs
```diff
@@ -401,13 +404,22 @@ private void CreateNameTextbox()
 
         private void CreateComboBreakDropdown()
```
Warp9000: rename to reflect new functionalty

Quaver.Shared/Screens/Selection/UI/Modifiers/Dialogs/Windows/JudgementWindowDialog.cs
```diff
@@ -401,13 +404,22 @@ private void CreateNameTextbox()
...
+
             ComboBreak = new JudgementWindowComboBreakDropdown(JudgementWindowsDatabaseCache.Selected)
             {
                 Parent = ContentBackground,
                 Alignment = Alignment.TopRight,
                 Y = NameTextbox.Y - 10,
                 X = -22
             };
+
+            LNMiss.Y = NameTextbox.Y + ComboBreak.Height;
```
Warp9000, line 422: why not move LNMiss initialization here instead of having it split?

Warp9000 (changes requested):

Quaver.Shared/Screens/Selection/UI/Modifiers/Dialogs/Windows/JudgementWindowDialog.cs
BakersDozenBagels: This is the only way I've found to set the Z-index correctly. If `LNMiss` is constructed after `ComboBreak` then `LNMiss` appears on top of `ComboBreak` when `ComboBreak` is open.

Quaver.Shared/Screens/Selection/UI/Modifiers/Dialogs/Windows/JudgementWindowDialog.cs
BakersDozenBagels: Changed to `CreateDropdowns()` in 04568bf33

Warp9000 (approved):

Issue #2861, opened 2021-02-20
Change ctrl+0 in song select from toggling pitch shifting with rates to resetting rate to 1


Illuminati-CRAZ:
![image](https://user-images.githubusercontent.com/42056854/108591577-4ba13000-731e-11eb-89bd-6552a36ce1d7.png)

Illuminati-CRAZ:
Reopening because yeah

<img width="595" height="737" alt="Image" src="https://github.com/user-attachments/assets/14af1bb5-9957-45d3-9de7-b7f29021c62b" />

PR #4513, opened 2025-08-16
Change ctrl+0 in song select from toggling pitch rate to resetting mods

Illuminati-CRAZ:
https://github.com/Quaver/Quaver/issues/2861

Decided to implement it as straight resetting mods because I think it makes more sense


AiAe:
Extended https://github.com/Quaver/Quaver/pull/4514

PR #4514, opened 2025-08-16
Add pitch toggle to be changed. Add new key bind to remove all mods

### LINKED ISSUE #2861: Change ctrl+0 in song select from toggling pitch shifting with rates to resetting rate to 1

AiAe:
We change the default pitch key toggle from `0` (Zero) to `\` (Slash) 
Remove all mods now uses `0` as default.

Both can be changed from Options, if someone wants to use 0 as their pitch toggle key.

Also closes https://github.com/Quaver/Quaver/issues/2861


Illuminati-CRAZ:
Now requires https://github.com/Quaver/Quaver.API/pull/206

Warp9000 (approved):

PR #4537, opened 2025-11-10
Add a button in timing group panel to select or cycle through the timing group(s) of the current selected notes

WilliamQiufeng:
https://github.com/user-attachments/assets/f2b3fba7-ca5b-4a50-9552-2da5067510ed


ESV-Sweetplum:
not sure this is necessary as plumogu beta already has this with keybinds

Warp9000 (approved):

PR #4544, opened 2025-12-13
Add a slider to tweak how much a long note is shrunk

### LINKED ISSUE #4455: Long Note Tail Slider.

It would be nice to have a way to reduce how far the long note tail. It would make it easier to read maps with very small releases. Obviously you would keep the hit timing for the long note the same, it would just visually make the long note smaller. I think this would be a nice addition since lots of tend to use percys from Osu!Mania and usually transfer over or play Quaver and Osu!Mania .

So maybe adding a slider or number to reduce how much of the long note tail is visible would be amazing, something like I've mocked up. 

Image

Stretching the sprites to get percys is really tedious, especially when you have to do with with 4 PNGs.

### Additional context

_No response_

WilliamQiufeng:
Resolves #4455 

Demo:
https://discord.com/channels/354206121386573824/398037301696921600/1449189489749528679

### DIFF
--- a/Quaver.Shared/Screens/Options/OptionsMenu.cs
+++ b/Quaver.Shared/Screens/Options/OptionsMenu.cs
@@ -182,6 +182,11 @@ namespace Quaver.Shared.Screens.Options
                     {
                         new OptionsSlider(containerRect, "Background Brightness", ConfigManager.BackgroundBrightness),
                     }),
+                    new OptionsSubcategory("Visuals", new List<OptionsItem>()
+                    {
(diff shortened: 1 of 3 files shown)


Quaver.Shared/Screens/Options/OptionsMenu.cs
7ef38c, line 1: Should probably add a tag for "Percy" as that's what some people will likely know it as

WilliamQiufeng: true

Warp9000 (approved):

PR #4561, opened 2025-12-29
Add tints to hit error chevron to reflect whether a hit is early/late or just in time.

WilliamQiufeng:
https://github.com/user-attachments/assets/341c9ffb-6904-48c4-a254-29b32b3f0cbd

* Added skinning options `HitErrorEarlyColor`, `HitErrorLateColor` and `HitErrorNeutralColor`: colors to represent early/late/correct
* Added option `HitErrorEarlyLateWindow`: the number of milliseconds of deviation to consider a hit as early/late instead of neutral

### DIFF
--- a/Quaver.Shared/Config/ConfigManager.cs
+++ b/Quaver.Shared/Config/ConfigManager.cs
@@ -510,6 +510,12 @@ namespace Quaver.Shared.Config
         /// </summary>
         internal static BindableInt HitErrorFadeTime { get; private set; }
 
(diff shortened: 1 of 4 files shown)


Janko5:
You beat me to a similar idea. I was planning to add icons representing whether you're clicking too fast/too late. Would you like to add that too? xd
https://streamable.com/tgqxe8

AiAe (approved):

PR #4609, opened 2026-05-16
Fix #4578: Timing Point Editor, Clamp BPM to a minimum of 0

### LINKED ISSUE #4578: Crash when setting BPM to a negative number
### Describe the bug

Untested on stable, but when setting a BPM to a negative number it crashes my game and it makes me sad :(

### To Reproduce

1. Go to the editor, and make a new chart
2. Set up a timing point
3. Lower the value to a value lower than 0

### Expected behavior

Game should not apply the BPM change

### Log Files

Couldn't get logs/went unlogged

### Screenshots

_No response_

2K-d:
# Fix #4578

## What I could reproduce:
When I'm in the Editor, in Timing Point Editor, and I make the BPM value go below 0 then Quaver crash.
Nothing logged inside 'runtime.log'.

## What I did to fix this:
I added a clamp of the BPM value to a minimum of 0 post-change.
I don't know much about what should be a correct BPM value but [0;+inf[ seemed to works so I kept it as is.
Maybe a [1;+inf[ range would be preferable?

### DIFF
--- a/Quaver.Shared/Screens/Edit/Plugins/Timing/EditorTimingPointPanel.cs
+++ b/Quaver.Shared/Screens/Edit/Plugins/Timing/EditorTimingPointPanel.cs
@@ -339,7 +339,7 @@ namespace Quaver.Shared.Screens.Edit.Plugins.Timing
 
             if (ImGui.InputFloat("##bpm", ref bpm, 1, 0.1f, format, ImGuiInputTextFlags.EnterReturnsTrue | ImGuiInputTextFlags.AutoSelectAll))
             {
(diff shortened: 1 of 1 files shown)


ESV-Sweetplum:
ideally set the minimum to 1

2K-d:
or maybe 0.1?

ESV-Sweetplum:
actually now that i think about it 0 should be fine, worst comes to worst you can set it to something like `1e-5`

WilliamQiufeng (approved):

PR #4610, opened 2026-05-17
Fix #4517: Don't consider workshop content without skin.ini as a skin

### LINKED ISSUE #4517: Steam Workshop Plugins Show Up in Skins Folder
### Describe the bug

# Overview 
When selecting a skin, the top section of the list is filled with random numbers, like so:

Image

# Why it's plugin related

Each of these corresponds to a subfolder in the `C:\Program Files (x86)\Steam\steamapps\workshop\content\980610` folder, which all contain plugins. Additionally, these folders all contain plugin.lua files. When selecting these "skins", the default skin is loaded. Clicking `Open Skin Folder` creates the warning:

> [!WARNING]
> Your skin folder doesn't exist.

### To Reproduce

1. Install any workshop plugin. The easiest one to see is [Pattern Master](https://steamcommunity.com/sharedfiles/filedetails/?id=2607285394&searchtext=patternmaster) because it contains a .py file.
2. Is this your card? No? Check under your skin

### Expected behavior

They shouldn't be there

### Log Files

not necessary, logless

### Screenshots

_No response_

2K-d:
# Fix 4517
## What I could reproduce
- I Installed Workshop plugin "Pattern Master"
- Found the plugin inside Skin selection
## What I have modified to fix this
If there is no "skin.ini" inside workshop content it should not be considered as a skin and should not be returned as a value in "GetSkins".

### DIFF
--- a/Quaver.Shared/Skinning/SkinStore.cs
+++ b/Quaver.Shared/Skinning/SkinStore.cs
@@ -947,8 +947,6 @@ namespace Quaver.Shared.Skinning
                         workshopList.Add($"Unknown <{new DirectoryInfo(directory).Name}>");
                     }
                 }
-                else
-                    workshopList.Add($"({new DirectoryInfo(directory).Name})");
             }
 
             workshopList.Sort();


ESV-Sweetplum:
fixes #4517

AiAe (approved):

PR #4612, opened 2026-05-20
Add a way to copy metadata's values across a whole mapset

A6sT:
This PR add a new button to the editor in the metadata's menu that will synchronize all of the value from one difficulty to the whole mapset. This button will not be visible unless there are at least 2 difficulties in this mapset.


AiAe:
@codex review

AiAe (changes requested):
Sync button should be moved with the other 2 buttons below.
When clicked, it should save, then sync and then close the modal.

AiAe (approved):

PR #4633, opened 2026-06-10
Song select menus optimizations

A6sT:
This aim to reduce / remove lagspikes when switching between map preview and other menus inside the map select screen.
Also tried to optimize leaderboard and option section, but can't really tell if this improves the performances


AiAe:
@codex review

AiAe (approved):

PR #4636, opened 2026-06-12
Chat messages overlapping: Round 2

A6sT:
Since my first fix didn't do the trick, i'm comming up with a new list of things that could cause the issue, that this PR aim to solve;
- Duplicate message -> The game now checks if a message already exists before sending it again (it's not just the message, it'll also check the time the message has been sent at)
- Message rendering -> I tried to change the message component behaviour so that they can't render on top of each other. This logic was already implemented before using the "TotalMessageHeight" variable, but i now recalculate the total height every time there is a new message sent. This is not perfect, but could potentially fix the issue

I also added a few optimizations to the way the chat messages are loaded when the message box is being initialized so that it won't tank fps on chat with a big history


AiAe:
@codex review

AiAe:
@codex review

AiAe (approved):

PR #4642, opened 2026-06-12
Fix lag in song select menu (v2)

A6sT:
One more small fix to make sure that there are things to write in the "Now Playing" temp file before actually writing inside


AiAe:
@codex review

AiAe (approved):

PR #4658, opened 2026-06-16
Implements Cocoa events to handle inputs so they are not bound to FPS on MacOS

AiAe:
Tested for awhile, did not find any issues so far.
Also added an toggle to disable the usage of cocoa.
It also now works in custom and limited only.


AiAe:
@codex review

PR #4669, opened 2026-06-21
Fix background flickering (second try)

A6sT:
This pr also include a fix to the playback speed dropdown menu from the editor that would display all of the options as being enabled


AiAe (approved):

AiAe:
Surely nothing will go wrong now right

PR #4676, opened 2026-06-23
[1.7.x] Add clan tag to chat messages

WilliamQiufeng:


AiAe:
Works great, but its missing displaying clan in Online Users
+ update submodule

AiAe:
Another issue I noticed while testing the PR, in multiplayer if you use the chat there it wont have the clan tag anymore, nor when you open the chat overlay.

AiAe (changes requested):
Quaver.Shared needs new compiled stripped lib, so people can run the game without the submodule.

AiAe (approved):

Issue #4595, opened 2026-04-21
Cards at "Active Downloads" get above buttons for other tabs


OrihsoArika:
### Describe the bug

Is just that. Yesterday, I downloaded a bunch of maps, my internet was kinda slow, then I decided to check the downloads and I saw that the cards get above the buttons for the other tabs.

### To Reproduce

1. Play at a wide resolution like 32:15(1280x600)
2. Download a bunch of maps.
3. While there's still maps that didn't download, try to scroll a bit
4. \*Visual bug happens*

### Log Files

[network.log](https://github.com/user-attachments/files/26929473/network.log)
[runtime.log](https://github.com/user-attachments/files/26929472/runtime.log)

### Screenshots

<img width="296" height="600" alt="Image" src="https://github.com/user-attachments/assets/643a7ed8-9ee4-4df4-95bd-ba1ef1d60394" />

A6sT:
This has been fixed with #4677

PR #4677, opened 2026-06-23
Import screen rewrite

A6sT:
This pr changes the importing screen to give more details about what's going on (include every stages of the importing process, a progress bar for each stage, the amount of map left to handle). It also affect the f5 screen (or force reload

This also potentially fix an issue (#4595) about download card overflowing and font size for theses cards


AiAe:
@codex review

Issue #4600, opened 2026-05-02
Quaver resetting monitor it displays on without prompt


krishangordhan:
### Describe the bug

Hey,

I find that when I am using Quaver it is automatically swapping to my other monitor when I select anything on the screen, it makes it hard to use the application when it swaps from the monitor I am using constantly.

Thanks,
Krishan

### To Reproduce

Windowed Quaver window

<img width="1012" height="542" alt="Image" src="https://github.com/user-attachments/assets/e0851ffe-1c44-4b40-9493-3bda87ba4c29" />
Settings posted.

Play a map
Quaver window swaps to monitor 1 from 2

### Expected behavior

When options are changed,
When a map is started
Quaver stays where it has been set on the screen before.

### Log Files

[runtime.log](https://github.com/user-attachments/files/27307538/runtime.log)
[network.log](https://github.com/user-attachments/files/27307537/network.log)

ESV-Sweetplum:
ensure that the monitor you want to play on is the "default monitor" in windows

krishangordhan:
I mean that's kinda what the bug is. It's changing to my default monitor without any prompts from me to do so. And I see no ability to set what monitor it should run on.

Issue #4505, opened 2025-08-03
resizing window resets some things


Nexility0:
### Describe the bug

2 issues in one maybe? (not sure if the second is intentional)

1st issue: resizing the game window while searching anything up in options removes the search 

2nd issue: resizing the game window resets the select screen back to the leaderboard instead of keeping the preview 

### To Reproduce

1st: 
1. go to options
2. type anything in the search bar
3. resize window

2nd: 
1. have preview open 
2. resize game window 

### Expected behavior

keep the search and preview open

### Log Files

[runtime.log](https://github.com/user-attachments/files/21563544/runtime.log)

### Screenshots

https://cdn.discordapp.com/attachments/1351727451184824373/1401385183189536808/Screen_Recording_2025-08-03_at_11.31.35_am.mov?ex=68901528&is=688ec3a8&hm=1b1dfa80522f287e6d0b790e4747aa60f5e5bab8d482c451f518467c9d4d265e& 

blame github not me, it wont lemme just put in the vid

2K-d:
The two cases come from the same root cause.
Inside 'QuaverGame.cs' there is a function 'ChangeResolution' called each time resize happens.
ChangeResolution always exits current screen to a new instance of the same screen.
The new instance have none of the information that old instance had.

ESV-Sweetplum:
also resizing window in skin editor discards all changes and forces you to return to main screen

Issue #4383, opened 2024-11-26
Maximized Window Slightly Offset


BakersDozenBagels:
### Describe the bug

Setting Quaver to windowed mode and maximizing the window offsets the visuals slightly. There is a transparent strip on the left side of the window, and the bottom and right are very slightly cut off.

### To Reproduce

1. Use Windows 11 Pro 23H2 22631.4460
2. Set Quaver to use a windowed display
3. Maximize the window with win+up

### Expected behavior

The window should fill the display area without being cut off.

### Log Files

N/A

### Screenshots

![image](https://github.com/user-attachments/assets/61ecc7bf-18ef-4d08-b657-fd04c13cfce3)
Notice Steam visible behind the window (on the left) and the spacing of the lower buttons and time remaining label.

PR #4679, opened 2026-06-24
[1.8.x] Fix weird offset+ dissapearance search in settings by changing resolution/window

TDMG-NL:
#### #4383 Fix window maximize offset in windowed mode

When Quaver is running in windowed mode and the window is maximized, the game viewport could become slightly misaligned, causing:
- A transparent strip on the left side of the window.
- Minor clipping on the bottom and right edges.

The viewport is now correctly resized to fill the available display area.

#### #4505 Preserve settings search when resizing the window

Resizing the game window while using the Settings search would clear the current search query.

The search state is now preserved during resize events.

####  #4600 Quaver resetting monitor it displays on without prompt

It now fully understands on which monitor the game is located and will change the resolution within that monitor

### Testing (Done on windows 11)

- Verified maximized windowed mode displays correctly without offsets or clipping.
- Verified Settings search remains active after window resize events.
- Verified window size can only be changed through the Settings menu options.
- Verified left tab options will stay open after resolution change
- Verified window stays on the current monitor after resolution change

### Closes

- #4383
- #4505
- #4600


AiAe:
Do we really have to fully disable the manual resize?
This is more of an MacOS issue but disabling that breaks some functionality for us, including going into "fullscreen" mode and also double clicking the window to make it full size and centered.

TDMG-NL:
"Do we really have to fully disable the manual resize?" no that does not have to, but A6st told me to make it that the player would need to use one of the options in the settings menu. just turning it back to `Window.AllowUserResizing = QuaverWindowManager.CanChangeResolutionOnScene;` will just give those functionalities back.

AiAe:
> "Do we really have to fully disable the manual resize?" no that does not have to, but A6st told me to make it that the player would need to use one of the options in the settings menu. just turning it back to `Window.AllowUserResizing = QuaverWindowManager.CanChangeResolutionOnScene;` will just give those functionalities back.

I think we for sure should revert the previous behaviour on how it worked in that regard, because right now this is kinda unusable for Mac specifically.
Alternatively we can also only allow this for the Mac platform, but I think some people on the other platforms might want this back as well.

TDMG-NL:
@AiAe There ya go!

AiAe:
Going from fullscreen, to moving the game to take half of the size breaks the rendering, until screen is changed:
CleanShot 2026-06-24 at 11  14 20

But from testing it also happens in previous versions, so maybe if this can also be fixed would be good?
If its not too much effort.

If its not happening on Windows, we can just ignore this probably.

TDMG-NL:
> Going from fullscreen, to moving the game to take half of the size breaks the rendering, until screen is changed: CleanShot 2026-06-24 at 11  14 20
> 
> But from testing it also happens in previous versions, so maybe if this can also be fixed would be good? If its not too much effort.
> 
> If its not happening on Windows, we can just ignore this probably.

On windows this is not happening. Could write that down for one of the later updates to the resolution logic when i work on the remaining issues about it, but i do not have a mac os system so it would be hard for me to confirm if my code will resolve the issue for mac users.

AiAe:
Sounds good, I asked Warp to test this PR on linux as well before I merge it.

TDMG-NL:
now includes the full fix on: https://github.com/Quaver/Quaver/issues/4505

TDMG-NL:
now includes the fix for #4600 and now this PR will fully close all issues about resolution changes. Testing still only has been done on windows so @AiAe and @Warp9000 when you guys have the time please check if it works properly. note that with this fix the game will boot on the primary monitor after restart

AiAe:
I'm not sure if in this PR we should also address the weirdness that i mentioned previously + that the leaderboard scrollbar does not update properly either when we change screen resolution

TDMG-NL:
> I'm not sure if in this PR we should also address the weirdness that i mentioned previously + that the leaderboard scrollbar does not update properly either when we change screen resolution

With "weirdness" do you mean the issue that's happening on macOS which you listed above? And about the leaderboard scrollbar, I can definitely look into that and update this PR with a fix for it

TDMG-NL:
It should be good to go now. @AiAe could you please check if the buffer helps with the problems macos has. i can't test that myself

PR #4684, opened 2026-06-24
[1.7.0] Fixed ln rendering issue with colorsnapping

A6sT:
It was because i forgot to remove a test i did to support colorsapping for ln body... my bad (v2)


A6sT:
Ok, actually, it wasn't me, but still because of me. This happened because we now use UseFallback = True on default skins, which caused weird interactions with ln base note when using colorsnapping. For now, the spritesheet implementation for ln have been removed, but will come back when we implement colorsnapping for ln bodies
Original pr that introduced the bug: #4665

A6sT:
for now, i'll just revert everything

Issue #3986, opened 2024-03-23
Switching to inaccessible leaderboards on ranked maps marking them as unranked 


Dopaminos:
### Describe the bug

video is the best explanation

https://github.com/Quaver/Quaver/assets/42616718/75d4448b-f0fc-403f-ae00-5fe09d20e6ea

### To Reproduce

reproducing steps is described in video

### Expected behavior

not marking them as unranked, switching to accessible (f/e global) leaderboards after scrolling away difficulty, inaccessible leaderboard was chosen on

### Log Files

not necessary

Dopaminos:
duplicate of https://github.com/Quaver/Quaver/issues/3979

Dopaminos:
nvm im dum oops

2K-d:
This reproduce in 1.7.0.21. But for some reason some map seems unaffected by this and other reproduce every time.

2K-d:
> This reproduce in 1.7.0.21. But for some reason some map seems unaffected by this and other reproduce every time.

Precision:
When map are sorted by title only map that have 1 difficulty get updated, if multiple difficulties exists on a map nothing get changed unless you update every single difficulty by selecting them with inaccessible leaderboard.
When map are sorted by difficulty every difficulty is getting updated.

2K-d:
If offline nothing is updated.

AiAe:
Should be fixed on canary

PR #4696, opened 2026-06-27
[1.8.x] fix-3986: Request on inaccessible leaderboards update Ranked Status to Unranked

2K-d:
When going inside the 'All' leaderboard a request is sent to update status of the map.
If your not a donator the result of the request contains a ranked status field that is always "unranked" even in case where it is incorrect.

What should happens I think is that an authenticated user that do not has correct role to access data should be met with a 403 (FORBIDDEN) status code and if so no update should be done on client.

I cannot access Quaver.Server.Client code so I cannot be sure of the status code currently used in those responses, but if it's not in [200;300[ range it should not be updated anyway.


2K-d:
fix #3986

AiAe:
Will close this PR, since we found an better solution which will soon be merged.

PR #4741, opened 2026-07-07
fix-4740: Humanizer change, use specific language library

### LINKED ISSUE #4740: Humanizer Filipino package broke the build
### Describe the bug

I'm opening this issue but I'm pretty sure there is nothing to do on the Quaver project except update Humanizer to a next update that fix this problem.
A similar issue is already created on the Humanizer project : https://github.com/Humanizr/Humanizer/issues/1804.

### To Reproduce

1. Clone the repo (current commit at the time I write this: bdfe09bfc96a54046454d107bb59904693d0cf4b)
2. dotnet run --project Quaver

### Expected behavior

It should build correctly without error

### Log Files

Will add supplementary log if anything is needed but most likely everything is already on the screenshot below.

### Screenshots

Image

2K-d:
Fix #4740
It is a temporary solution while waiting for an official fix from the library


AiAe:
I also noticed that this library is creating a lot of language folders in the project? Or it might be also some other side effect of something I changed.
Gotta look into that first.

AiAe:
@2K-d so whats the reason for adding 2 languages specifically to be included?

2K-d:
It's not 2 added, it's one removed "Humanizer", that contains every language, and 3 added "Humanizer.Core" "Humanizer.Core.fr" "Humanizer.Core.zh-CN" (the 3 languages that were fully translated at the time I write this but you can add more). The important part is to not include "Humanizer.Core.fil" (which is in "Humanizer") because it breaks compilation for some unknown reason (an issue was already openned on Humanizer project before it was used as a dependency here).

AiAe:
I see, will check it out soon then

PR #4744, opened 2026-07-07
New code based buttons

TDMG-NL:
only works for the buttons in options right now. if working alright will update this PR with the implementation on all buttons


TDMG-NL:
check the ones in the options menu. i feel like they are too big

AiAe:
> check the ones in the options menu. i feel like they are too big

Lower them by -2, I'm also doing that in my other PR.

PR #4745, opened 2026-07-08
Added a temporary function to remove the overlapping keybinds for mines

A6sT:


WilliamQiufeng (commented):
I think there may be a better solution, cuz right now you are basically banning R as the resnap action. What about when a keybind with conflicting actions are triggered a warning is shown?

PR #4739, opened 2026-07-06
Remove ImageSharp dependecies and optimize banner loadings in song select screen

A6sT:
hopefully this will improve performances

PR #4752, opened 2026-07-11
Follow up to banner optimizations issues

A6sT:
This is a follow up to the issues mentionned in #4739 that were not fixed when merged


AiAe:
@codex review

PR #4757, opened 2026-07-12
Move mine hit in place of scores

WilliamQiufeng:


AiAe:
I think we wont be touching 1.7 anymore so i will close it

PR #4810, opened 2026-07-15
Editor improvements 1

AiAe:
Waveform/Spectrogram still renders normally and its accurate - i made it so its loading in chunks and only display the needed chunks (useful for long maps)

Seekbar/diff bar - is now single sprite, check if u notice any difference vs old one

SV/SF lines - are also now single sprite, check if they are accurate

Exiting editor now is faster and does not tank FPS

---

Might end up splitting this into multiple prs, but for now it will be in testing phase for awhile


AiAe:
Closing this one, because I opened dedicated pull requests for each section we improved for better readability

PR #4811, opened 2026-07-15
[1.8.x] Add Tournament Playlists

A6sT:
This PR implements [#4786](https://github.com/Quaver/Quaver/discussions/4786) as Tournament playlists, and also revert some local leaderboard changes and fixes some font size in gameplay

# What does Tournament playlist do ?
- It adds tournament playlist: creating / copying a playlist asks for a type to create / copy it as (so normal and tournament)
- When maps are added to a tournament playlist, you can right click on them to "save current mods" for that map. This will then change the difficulty of the map to match the modifiers used (so, if fln is picked for example, a 20 diff map won't be displayed as 20 but at the difficulty calculated by the fln mod). This can be seen from the playlist views when looking at lowest / highest maps difficulty
- Import / export of playlist now includes type. When importing a tournament playlist and when you're not the owner of it, you won't be able to change mods, except for mirror and no miss


AiAe:
@codex review

PR #4822, opened 2026-07-17
Fix map deletion hiding all mapsets

A6sT:


AiAe:
This does not resolve the bug for me, when I delete mapset the selection screen still goes blank

AiAe (changes requested):
Still does not work when deleting mapset

AiAe (approved):

PR #4827, opened 2026-07-19
Fix Long Notes selection being missplaced after transforming into a normal note

A6sT:
To reproduce (thanks to Larpzoid):
step 1: add long note (set it to any length)
step 2: drag ln tail to remove ln
step 3: select said note
step 4: contemplate what happens when life doesn't give you lemon


hexyll:
fuck it i guess im officially larpzoid now and forever :waga:

AiAe (approved):

PR #4840, opened 2026-07-20
Fix editor flagging false manual edits

A6sT:
This PR also contains an issue where the game would crash when reloading the editor after switching audio file reference in the .qua file


AiAe:
@codex review

PR #4916, opened 2026-08-23
Hover focus result screen + fix for judgement graph lines showing default windows instead of the score's custom judgement windows

TDMG-NL:
this pr creates a simple highlighting system when hovering over judgements, mean, mine miss or std.dev so its easier to see what happened where. There is also a zone between the mean and std.dev so you can see both at the same time.


TDMG-NL:
Added a fix for judgement graph lines showing default windows instead of the score's custom judgement windows when checking a score that has been played with a custom preset

AiAe:
1. The hovering feels kinda clunky when there is no animation/active/hover state
2. Its weird to still see the others dots even when faded, maybe best to fully hide them?
3. With making the lines fully colored, they sometimes will hide some of the dots from the scatter when they are on the line exactly

A6sT:
2. was my suggestion, it feels really weird to see all timings vanish just because you hovered some ""random"" part of the result screen. Might have to do with the fact that the transition feels clunky as you said though

AiAe:
> 2. was my suggestion, it feels really weird to see all timings vanish just because you hovered some ""random"" part of the result screen. Might have to do with the fact that the transition feels clunky as you said though

Perhaps having something as a hover change will be plenty on telling the user whats going on

PR #4921, opened 2026-08-24
Map offset could be changed while in multi lobby

TDMG-NL:
this pr "fixes" it so like before you can only change the map local offset while ingame. this because it was now overlapping inputs like changing the speed rate

### DIFF
--- a/Quaver.Shared/Screens/Multi/MultiplayerGameScreen.cs
+++ b/Quaver.Shared/Screens/Multi/MultiplayerGameScreen.cs
@@ -301,13 +301,6 @@ namespace Quaver.Shared.Screens.Multi
                 return GlobalInputHandleResult.Consumed;
             }
 
-            if ((action.BaseWithLayer()) == GlobalKeybindActions.IncreaseOffset &&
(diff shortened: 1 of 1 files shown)


WilliamQiufeng:
I feel like it would be better if it could be tweaked from within the lobby. There's a lot of times where I want to do that before I start, because it is clear from preview what offset it has.

TDMG-NL:
@WilliamQiufeng might it not be better to also implement that for the song select? This pr was Made to simply make those the same so I guess it would be fine (if not the same keybind) as long as we keep it consistent

WilliamQiufeng:
> @WilliamQiufeng might it not be better to also implement that for the song select? This pr was Made to simply make those the same so I guess it would be fine (if not the same keybind) as long as we keep it consistent

The issue with song select is it has that search bar. Might need to think about how the interaction works

TDMG-NL:
> The issue with song select is it has that search bar. Might need to think about how the interaction works

For sure agree, but I personally think we do need to keep it consistent. Atleast i think its weird to have different interactions in song select and multi lobby

PR #4927, opened 2026-09-05
Add playback control to plugin state #2 (honk shoo honk shoo)

ESV-Sweetplum:
Test plugin: [temp.zip](https://github.com/user-attachments/files/31869488/temp.zip)

Adds new field, `state.IsPlaying`, which returns a boolean corresponding to the editor playback status.
Now allows `state.SongTime` to be set, which allows the plugin developer to skip `actions.GoToObjects` regex manipulation.
Added two new functions, `actions.StartPlayback()` and `actions.StopPlayback()`. Both of these functions were tested to make sure that calling the same one twice does not cause any issues.

absolutely no other changes were performed at any moment in time

### DIFF
--- a/Quaver.Shared/Screens/Edit/Actions/EditorPluginActionManager.cs
+++ b/Quaver.Shared/Screens/Edit/Actions/EditorPluginActionManager.cs
@@ -209,6 +209,14 @@ namespace Quaver.Shared.Screens.Edit.Actions
 
         /// <summary>
         /// </summary>
+        public void StartPlayback() => ActionManager.StartPlayback();
+
+        /// <summary>
+        /// </summary>
+        public void StopPlayback() => ActionManager.StopPlayback();
+
+        /// <summary>
+        /// </summary>
         /// <param name="input"></param>
         public void GoToObjects(string input) => ActionManager.GoToObjects(input, true);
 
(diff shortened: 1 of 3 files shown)


Quaver.Shared/Screens/Edit/Actions/EditorPluginActionManager.cs
```diff
@@ -207,6 +207,14 @@ public void RemoveScrollSpeedFactorBatch(List<ScrollSpeedFactorInfo> sfs, Scroll
         /// <param name="tp"></param>
         public void ResetTimingPoint(TimingPointInfo tp) => ActionManager.ResetTimingPoint(tp, true);
 
+        /// <summary>
+        /// </summary>
+        public void StartPlayback() => ActionManager.StartPlayback();
+
+        /// <summary>
+        /// </summary>
+        public void StopPlayback() => ActionManager.StopPlayback();
```
WilliamQiufeng: You made a `fromLua` parameter on those methods. Could you pass a `true` in there, or remove that parameter?

ESV-Sweetplum: can do

WilliamQiufeng (approved):

